PR #12, opened 2025-11-27
Core classes discussion

ollmer:
This is very high-level first draft of the core classes described in our [Excalidraw system diagram](https://excalidraw.com/#room=6b2ae51f15115f830ccf,6luinCiMcoiexaJHmObfAQ).

Main sections are:
1. Action abstractions
2. LLM interaction abstractions
3. Env, Benchmark and Task abstractions
4. Agent abstraction
5. Single run abstraction, contains main loop

Please share your comments here, preferrably attributed to a specific lines in a markdown document. After some review period we will establish a call to discuss all the comment.

### DIFF
--- a/docs/core_classes.markdown
+++ b/docs/core_classes.markdown
@@ -68,8 +68,9 @@ class Observation(BaseModel):
         pass
 ```
 
-## LLM interaction abstractions, LiteLLM based:
-[https://docs.litellm.ai/docs/completion/input](https://docs.litellm.ai/docs/completion/input)
+## LLM interaction abstractions
+Based on LiteLLM conventions:  
+[https://docs.litellm.ai/docs/completion/input](https://docs.litellm.ai/docs/completion/input)  
 [https://docs.litellm.ai/docs/completion/output](https://docs.litellm.ai/docs/completion/output)
 ```
 class LLMMessage(BaseModel):


docs/core_classes.markdown
```diff
@@ -0,0 +1,228 @@
...
+        arguments (Any): The arguments to be passed to the function.
+    """
+    id: str
+    name: str
+    arguments: Dict[str, Any]
+
+class Content(BaseModel):
+    """Represents a piece of content in an observation."""
+    type: str  # e.g., "text/plain", "image/png"
+    data: Any  # The actual content data
```
recursix: We need to make sure we can support:
* introspection from generalist agent and that it can format automatically to litellm API
* wide range of format e.g. image, video, sound, even potentially (in the future) remote sensing data e.g. multi-spectral hyperspectral

I'm wondering if the Content should have derived classes for the different option or if `data: Any` should have different classes of data content.

docs/core_classes.markdown
```diff
@@ -0,0 +1,228 @@
...
+
+class Observation(BaseModel):
+    tool_call_id: str | None = None # first observation may not be linked to any tool call
+    contents: list[Content]
+    metadata: dict = Field(default_factory=dict)
+    reward_info: dict = Field(default_factory=dict)
+
+    def to_messages(self) -> List[LLMMessage]:
```
recursix: I think this should just loop over the content and call to_message on the different content.

docs/assets/images/classes_schema.png
manuel-delverme: Do we need the task to communicate directly to the agent? reset()/first observation usually defines the goals

docs/core_classes.markdown
```diff
@@ -0,0 +1,228 @@
...
+```
+class LLMMessage(BaseModel):
+    """Represents a message for the LLM input"""
+    role: Literal["system", "user", "assistant", "function", "tool"]
+    content: str | list[Dict[str, Any]] | None #  may be null for assistant messages with function calls.
+    name: Optional[str] = None  #  required if the role is "function". The name should match the name of the function represented in the content
+    function_call: Optional[ToolCall] = None  #  may be present if the assistant is calling a function
+    tool_call_id: Optional[str] = None  #  may be present if the message is from a tool result
+
+class Prompt(BaseModel):
```
recursix: We may call that `Discussion` or `Conversation` since it's a list of messages i.e. prompt is too generic

In terms of modeling, we have to take into account various things.
* this may be a growing conversation since the beginning of the episode, or since the begining of the step, relaying on some summary of the episode
* it's important to have caching well integrated and cacing mechanisms works differently for different providers
* an agent may call different LLMs within a single step could be sequential tool call, with a growing conversation, but could be e.g. planner + grounder, which would be 2 different LLMs and have different history. 

The two main reasons to gather this information is:
* For visualization
* For Post-training

docs/core_classes.markdown
```diff
@@ -0,0 +1,228 @@
...
+    id: str
+    name: str
+    arguments: Dict[str, Any]
+
+class Content(BaseModel):
+    """Represents a piece of content in an observation."""
+    type: str  # e.g., "text/plain", "image/png"
+    data: Any  # The actual content data
+
+class Observation(BaseModel):
```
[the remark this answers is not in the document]
ollmer: I would say that `Observation` object is exactly `o` in this notation, and its POMDP-style partial observability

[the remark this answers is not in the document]
ollmer: The source of the reward is also a little more complex matter in RL than in the classic definition. We could think of:   
a) task being source of the reward, but it could be conditioned on almost any info available at the moment, including env internal state and whole episode so far. So not only `f(s, a, s')`, but more like `f(s, tau)`
b) env being the source of the reward, where `r = f(s)`
c) judges! This part is still missing in the core and we need to define them. Judge call could also be conditioned on any info, so `r_estimate = judge(s, tau)`

docs/core_classes.markdown
```diff
@@ -0,0 +1,228 @@
...
+
+    def close(self):
+        pass
+
+class Task(BaseModel):
+    id: str
+    evaluate_per_step: bool = False
+
+    def setup(self, environment: Environment) -> list[Observation], dict:
+    """ Returns goal obsevation, initial observation and a dict with additional task info"""
```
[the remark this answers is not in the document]
conte91: > Or same for CUA where you can reuse same docker-based env.

(Uninformed opinion, sorry if i'm misunderstanding) This is the environment's concrete implementation though, which can be easily shared, and is not the same as the "environment" in the RL sense. For example, suppose we have a docker "env" (for example for us, a docker container serving a mocked ERP) that we want to share across tasks. Each task can still have its own `Environment`:

```
class ERPEnvironment:
   def __init__(self, erp_docker: DockerInstance) -> None:
       # whatever, just an example

# (And two tasks that run on `ERPEnvironment`) 
def __main__():
    erp = docker.run(something)
    benchmark.run([ERPTask1(ERPEnvironment(erp)), ERPTask2(ERPEnvironment(erp))])
```


The two `Environment` objects can still operate independently as @manuel-delverme suggested, while the environment's dependency is shared

docs/core_classes.markdown
```diff
@@ -0,0 +1,228 @@
...
+        pass
+
+    def teardown(self, environment: Environment):
+        pass
+
+    def validate(self, environment: Environment, trace: Trace) -> dict:
```
conte91: Here and elsewhere: this whole generic structure is the same that we currently have in Browsergym and gives raise to a major issue, where the user is tricked into believing there's a (false) genericity of the Task and Environment, while IRL the two are tightly coupled together. If this task is `ReadEmailFromOutlookTask`, I can't tell it to `task.validate(GMailEnvironment)`.

In practice, all these functions for all the task will start with
```
if not isinstance(environment, WhateverEnvironmentIActuallyNeed):
   raise ValueError("unsupported")
```

Because in order to validate the task, it needs to know _something_ about the specifics of the environments it's targeting. 

One solution to this problem would be to make `Task` a generic on an `Environment` protocol.  So that:

```
class OutlookEnvironment(ABC):
    # For example
    def get_unread_email_count():
        ...


class ReadOneEmailTask(Task[OutlookEnvironment]):
    def validate(self, environment: OutlookEnvironment, trace: Trace) -> dict: # See other comment on the dict
         return {"success": environment.get_unread_email_count == 0} # now environment is well-defined!
```

conte91: Here and in the `Observation` class: similar to the `Environment`-`Task` pair, returning a `dict[Any, Any]` gives the user no guidance on what a task should return. if this was the case, it'd be amazing - but in practice, we can't afford that, because whatever loop function or benchmark runner we use will need to know that there is (for example), a "success" or "done" or "reward" field in here.

The two ways in which this information will be propagated are word-of-mouth and insider knowledge (i.e. you check out other `Task`s and notice that the kosher way of implementing `validate` is to return a `dict` which contains a `"success"` bool key). So there's again a false sense of generality communicated to the user, that doesn't match the real life (as in, "you can put anything here as long as it's the correct thing").

One way to circumvent this limitation is to provide structured dataclasses/pydantics for the various observation types (I believe ADP does exactly this for the Observation class) and leave a `metadata`/`extra` field for all those things that are _never_ used by this codebase but the user might want to save for themselves. 

My point basically is that:
- Every data that is _required_ by our codebase/infra should be _explicitly_ defined in the interfaces - so that it's obvious for users what they need to type where (and it's easier to document)!
- All the "generic" types should have a mixture of "all the things AL2 _requires_ to work" (e.g. in this case, the "success" and "done" fields) and a "extra" dict that is for the user to propagate whatever they want.
- Our codebase should _never_ try to access the `extra` dict, since no assumption can be made about its contents
    - (optionally) We can actually enforce this if the internal classes always receive the top-level (only required fields) pydantic in their typing, and the extra is relegated to a public-facing model.

docs/assets/images/classes_schema.png
manuel-delverme: >env.reset() and task.setup() should be called by the main loop code

The standard RL interface is:
```python
obs = env.reset()
while not done:
    action = agent.act(obs)
    obs, reward, done, info = env.step(action)
```

Most users will expect this. Adding a separate task.setup() that the main loop must orchestrate is non-standard complexity.

Suggestion: What we call "Benchmark" should expose the standard Environment interface to the agent:

```python
  class Environment:  # What the agent sees (currently called Benchmark-taks)
      def __init__(simulator, task):
          self.simulator = simulator
          self.task = task
          
      def reset(self) -> Observation:
          self.simulator.reset() # Reset transition dynamics
          return self.task.setup(self.simulator) # Setup goal, return initial obs

      def step(self, action) -> Observation:
          return self.simulator.step(action)
```

The internal split between "Environment" (transition dynamics) and "Task" (goal/reward definition) is an implementation detail the agent shouldn't see.

This preserves internal separation (reusable envs, task-filtered actions) while giving agents a familiar interface.

Naming proposal:
`Environment`: what the agent interacts with (the RL "env")
`Simulator`: the transition dynamics (browser, vm, tools, docker, etc.)
`Task`: goal definition, reward, validation (internal to Environment), judges live here

This preserves internal separation for reuse (same BrowserSimulator across tasks) while giving agents the familiar RL interface. The split between simulator and task becomes an implementation detail.

docs/core_classes.markdown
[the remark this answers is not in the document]
manuel-delverme: >f(s, tau)

tau should not be necessary if we have internal state but i see why we might want to keep it in practice, e.g. to answer questions like: was the agent being efficient/safe more easily, even if looking at the observations/actions could create hard to spot bugs where the agent perception is correct while the internal simulator dynamics are incorrect (we rely on features to represent the state correctly, instead of internal state).

On reward sources: I'm not sure having rewards come from multiple sources (task, env, judge) is good design.
Looking at the current sketch:

line 64: Observation has reward_info
class Observation(BaseModel):
    reward_info: dict = Field(default_factory=dict)

line 220-222: Task populates it
if self.task.validate_per_step:
    self.task.validate_step(self.environment, llm_output.tool_calls, observations)

My suggestion: Environment (or Simulator in the other comment) should only be the transition function, it surfaces raw state changes (including crashes, errors, timeouts) but assigns no reward to them. The Task/Judge decides:
1. Whether a crash is terminal or recoverable
1. Whether an error deserves negative reward
1. What constitutes "success"

This means:
Environment.step() returns Observation with raw contents only (no reward_info)
Task.validate_step() / Judge interprets observations and produces rewards
An "online judge" is really just a Task with evaluate_per_step=True and humans/llms being invoked.

docs/core_classes.markdown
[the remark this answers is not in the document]
manuel-delverme: See my other comment:
https://github.com/The-AI-Alliance/AgentLab2/pull/12/files#r2600856862

manuel-delverme (commented):
The comment of rotating names:

| **Before**   | **After**     |
|--------------|----------------|
| Environment  | Simulator      |
| Benchmark    | Environment    |

would clarify things a lot

docs/core_classes.markdown
```diff
@@ -0,0 +1,228 @@
...
+    def actions(self) -> List[Tool]:
+    """ Returns list of actions supported by that environment"""
+        pass
+
+    def step(self, tool_call: ToolCall) -> Observation:
+        pass
+
+    def multi_step(self, tool_calls: List[ToolCall]) -> List[Observation]:
+        """ Execute multiple function calls in parallel"""
+        pass
```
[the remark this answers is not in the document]
ollmer: what is the order of trajectory in case of composite [click, fill]? My understanding is for the strictly sequential envs (which would be the majority of envs) it should end up with [click, obs_after_click, fill, obs_after_fill] and not [[click, fill], [obs_after_click, obs_after_fill]] as it would be in case of composite actions and obs.
For the case with env that performs parallel excution (for example that includes multiple tools with separate apis like calculator and browser) it would be [[click, fill], {obs_after_click, obs_after_fill}] where the order of click and fill is defined by the llm output order and the order of observations is not defined due to concurrent origin

PR #21, opened 2025-11-28
Add ruff format/linter and initial dir structure

amanjaiswal73892:
Adds bare package directory and configs.

**Project setup and configuration:**

* Added `pyproject.toml` with project metadata, Python version requirement (>=3.13), build system configuration, and Ruff formatter/linter settings. Also defined a development dependency group including Ruff.
* Added `.python-version` file specifying Python 3.13 for environment consistency.

**Development tooling:**

* Added `.vscode/settings.json` to configure Ruff as the default Python formatter and enable format-on-save for Python files in VS Code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,18 @@ make lint      # Lint and auto-fix
 make help      # Show all commands
 ```
 
+## Project Structure
+
+```
+AgentLab2/
+├── src/agentlab2/       # Source code for the framework
+├── tests/               # Test suite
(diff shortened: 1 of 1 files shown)


ollmer (changes requested):
a few thoughts on formatter settings and we're good to go when settled.

ollmer (approved):
LGTM!

ollmer:
@amanjaiswal73892 when do you want to merge it?

PR #32, opened 2025-12-10
OTEL-native metrics tracing with automatic episode export

### LINKED ISSUE #24: Telemetry / Data format and storage

manuel-delverme:
Adds a tracing system built on OpenTelemetry SDK for recording agent episodes.

```
AgentTracer API
      │
      ▼
 OTEL TracerProvider
      │
      ├──► DiskSpanExporter ──► traces.jsonl (all spans)
      │           │
      │           └──► TraceProcessor (async) ──► experiment/episode/step_{N}.json + .zip
      │
      └──► OTLPSpanExporter (optional) ──► remote collector
```

### Usage
```python
tracer = AgentTracer(service_name="my-agent", run_dir="./runs")

with tracer.benchmark("experiment_1"):
    with tracer.episode("episode_0"):
        tracer.log({"action": "click", "reward": 0.5})

tracer.shutdown()
```

### Output
```
runs/
  traces.jsonl              # All spans (OTEL format)
  experiment_1/
    episode_0.zip           # Auto-zipped on episode end
      episode_0/
        step_0.json
        step_1.json
```

### Key decisions
- **OTEL-native**: Works with Phoenix, Jaeger, any OTLP collector
- **Async episode export**: Non-blocking via ThreadPoolExecutor
- **Backpressure**: Warns + flushes at 10 pending exports
- **JSONL storage**: Human-readable, append-only, crash-safe

### Files
| File | Purpose |
|------|---------|
| `tracer.py` | Public API: `benchmark()`, `episode()`, `step()`, `log()` |
| `disk_exporter.py` | OTEL exporter → JSONL + async episode zip |
| `store.py` | Thread-safe JSONL writer |
| `processor.py` | Episode → JSON files → zip |
| `models.py` | `SpanRecord` pydantic model |

Closes #24


manuel-delverme:
Output of the metrics:
 
image

recipes/hello_world_metrics.py
```diff
@@ -0,0 +1,53 @@
...
+        output_dir="./hello_world_1",
+        env_config=env_config,
+        agent_config=agent_config,
+        benchmark=benchmark,
+    )
+
+    tracer = AgentTracer(service_name=exp.name, output_dir=exp.output_dir)
+    logging.info(f"Metrics run_id={tracer.run_id}, output={tracer.output_dir}")
+
+    try:
```
ollmer: can you please update the example to showcase how to properly use tracer with `experiment.run_ray()`? It would be the most popular scenario, as a sequential eval only intended to use for a debug purpose.

ollmer:
Yep, transport between main process and children through env vars looks ugly. We're already passing experiment name to a child explicitly as a part of its arguments, could we reuse it in every child process? So we have there a code like:
```
with tracer.benchmark(config.exp_name):
    with tracer.episode(config.episode):
```
in every child process?

manuel-delverme:
> Yep, transport between main process and children through env vars looks ugly. We're already passing experiment name to a child explicitly as a part of its arguments, could we reuse it in every child process? So we have there a code like:
> 
> ```
> with tracer.benchmark(config.exp_name):
>     with tracer.episode(config.episode):
> ```
> 
> in every child process?
Not sure about:
config.episode

Ehe experiment/benchmark is configurable, the episode is just a sub element of that i would rather the config be frozen and have an extra variable object

PR #271, opened 2026-04-08
RFC: BrowserGym tool refactor plan (not for merge)

recursix:
## Not for merge -- discussion only

This PR contains a plan doc (`docs/BGYM_TOOL_REFACTOR.md`) for refactoring `BrowsergymTool` to be a thin wrapper over bgym's native action/observation space.

**Please comment directly on the .md file or in this PR.**

### Key changes proposed:
- Delete `BidBrowserActionSpace` / `BrowserActionSpace` protocols
- Build `action_set` dynamically from `HighLevelActionSet.to_tool_description()`
- Override `execute_action()` to serialize Action back to bgym action string
- Make action subsets configurable per benchmark ("workarena", "webarena", etc.)
- Surface `send_msg_to_user` and `report_infeasible` errors to the agent (builds on #270)

### Open questions (in the doc):
1. Should `send_msg_to_user` be exposed as an agent-callable action?
2. Can we drop the checkbox JS fallback if Playwright fixed the bug upstream?
3. Enable bgym's `multiaction=True`?

Prereq: #270 (error propagation fix)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #270, opened 2026-04-08
fix(browsergym): surface infeasible action errors and last_action_error

recursix:
## Summary

- **Bug 1**: `report_infeasible_instructions` callback in `_execute_bgym_step` was silently logging a warning — the agent received `"Success"` even when bgym reported an action as infeasible (e.g. bid doesn't exist). Now captured and returned as `"Failed (infeasible): ..."`.
- **Bug 2**: `last_action_error` was checked in `_bgym_obs_to_cube_obs` but never populated by `_extract_bgym_obs`, making it dead code. Now populated from `_last_info["action_error"]`.

## Context

When comparing cube-harness's `BrowsergymTool._execute_bgym_step()` against browsergym's native `BrowserEnv.step()`, these two error propagation paths were missing. This likely explains why agents never see bgym-level errors during runs.

## Test plan

- [ ] Run a MiniWoB or WorkArena task that triggers an invalid BID — verify agent receives `"Failed (infeasible): ..."` instead of `"Success"`
- [ ] Verify `last_action_error` appears in observation contents after a failed action

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #272, opened 2026-04-08
refactor(browsergym): thin wrapper over bgym action/observation space

recursix:
## Summary

- **Delete `BidBrowserActionSpace` / `BrowserActionSpace` protocol layer** — removed `action_spaces/` directory entirely
- **Dynamic `action_set`** — built from `HighLevelActionSet.to_tool_description()`, agent sees bgym's native action names (`click`, `fill`, `scroll`, `send_msg_to_user`, ...) with full parameter schemas
- **`execute_action()` override** — serialises `Action` back to bgym action string, executes via `to_python_code()` + `execute_python_code()`
- **Configurable `action_subsets`** — set per benchmark (`"workarena"`, `"webarena"`, `["bid", "nav", "tab", "chat", "infeas"]`, etc.)
- **Error propagation fixed** — `report_infeasible_instructions` and `send_msg_to_user` callbacks captured and surfaced to agent (supersedes #270)
- **`last_action_error`** populated in observation dict
- Checkbox/radio JS fallback preserved as internal hook
- Net **-354 lines** (333 added, 687 removed)

## Context

See RFC discussion in #271. The old `BidBrowserActionSpace` protocol defined 12 cube-specific method names that manually mapped to bgym action strings, hiding bgym's native names and missing actions (scroll, dblclick, focus, clear, upload_file, tab ops). This PR replaces that with direct exposure of bgym's `HighLevelActionSet`.

## Test plan

- [x] All 657 existing tests pass
- [x] New tests for `_action_to_bgym_string` serialisation (15 cases)
- [x] New tests for `_build_action_schemas` (3 cases)
- [x] New tests for infeasible/user_message capture
- [x] New tests for configurable action subsets
- [ ] Manual: run a WorkArena task and verify agent sees bgym native action names


recipes/allac_recipe.py
younik, line 1: We shouldn't merge this, isn't it?

cubes/workarena/src/workarena_cube/task.py
```diff
@@ -221,15 +220,26 @@ def close(self) -> None:
...
-        meta = WorkArenaBenchmark.task_metadata[self.task_id]
+        if self.task_class_path is None:
+            raise ValueError(f"task_class_path is required to instantiate WorkArenaTask (task_id={self.task_id})")
+        meta = TaskMetadata(
+            id=self.task_id,
+            extra_info={
+                "task_class_path": self.task_class_path,
+                "requires_chat_answer": self.requires_chat_answer,
+            },
+        )
```
younik: Is this the workflow we want? cc @NicolasAG 
If I remember correctly, we wanted to have a single instance of the TaskMetadata, and this is already in the benchmark
This (1) duplicates the metadata instance (2) it is created in two places, potentially diverging

cubes/workarena/src/workarena_cube/task.py
```diff
@@ -198,12 +190,19 @@ def finished(self, obs: Observation) -> bool:
...
+          agents from using it as a reasoning channel and wasting their action budget.
+        """
+        filtered = actions
         if self._chat_tool is None:
-            supported_actions = supported_actions - {"send_message"}
-        filtered = [a for a in actions if a.name in supported_actions]
-        logger.debug(f"Filtered {len(filtered)} out of {len(actions)} actions for WorkArena task.")
+            filtered = [a for a in filtered if a.name != "send_message"]
+        if not self.metadata.extra_info.get("requires_chat_answer", False):
+            filtered = [a for a in filtered if a.name != "send_msg_to_user"]
```
younik: ? What is the difference between send_message and send_msg_to_user?

cubes/workarena/src/workarena_cube/task.py
```diff
@@ -177,9 +161,17 @@ def reset(self) -> tuple[Observation, dict[str, Any]]:
...
+        Prefers ChatTool messages when present. Falls back to the browser tool's
+        chat_messages property (populated via send_msg_to_user when using BrowsergymTool).
+        Tools without chat_messages (e.g. SyncPlaywrightTool) return an empty list.
+        """
         chat = self._chat_tool
-        return chat.messages if chat is not None else []
+        if chat is not None:
+            return chat.messages
+        tool = self._browser_tool
+        return getattr(tool, "chat_messages", [])
```
younik: why the browser tool should have chat_messages?

cubes/workarena/src/workarena_cube/task.py
```diff
@@ -198,12 +190,19 @@ def finished(self, obs: Observation) -> bool:
...
+
+        - send_message: only available when a ChatTool is present
+        - send_msg_to_user: only available for tasks whose validate() reads chat_messages
+          (i.e. tasks that expect a natural-language answer from the agent, like chart
+          value retrieval or knowledge-base search). Hiding it from other tasks prevents
+          agents from using it as a reasoning channel and wasting their action budget.
+        """
+        filtered = actions
         if self._chat_tool is None:
-            supported_actions = supported_actions - {"send_message"}
```
younik, line 204: why we don't filter anymore for supported action? Is this a common strategy for cubes, to let all the tools action be available? Then what is the point of filter_actions function? @NicolasAG

cubes/workarena/src/workarena_cube/benchmark.py
```diff
@@ -66,12 +67,21 @@ def _setup(self) -> None:
...
         self._runtime_context = {"level": self.level, "n_tasks": len(task_tuples)}
         metadata: dict[str, TaskMetadata] = {}
+        seen_classes: set[str] = set()
         for task_class, _seed in task_tuples:
             task_id = task_class.get_task_id()
             task_class_path = f"{task_class.__module__}.{task_class.__qualname__}"
+            if task_class_path not in seen_classes:
+                seen_classes.add(task_class_path)
+                validate_src = inspect.getsource(task_class.validate)
+                requires_chat_answer = "chat_messages[-1]" in validate_src
```
younik: I think we shouldn't decide if send_message action exists or not, based on the task, especially if this doesn't expose it nicely. The agent can always have the send_message action, regardless if it is needed or not to complete the task. This is how it is done for browsergym, so if you changed this to improve the perf, it should not be the right path

src/cube_harness/tools/browsergym.py
```diff
@@ -38,48 +37,81 @@ class BrowsergymConfig(ToolConfig):
...
         self._last_info: dict | None = None
         self._last_reward: float = 0.0
         self._last_terminated: bool = False
+        # Accumulated send_msg_to_user messages — passed to WorkArena validate() as chat history
+        self._chat_messages: list[dict] = []
+
+    @property
+    def chat_messages(self) -> list[dict]:
+        """Messages sent via send_msg_to_user, in bgym chat protocol format."""
+        return self._chat_messages
```
younik: chat message shouldn't be part of a browser tool

src/cube_harness/tools/browsergym.py
```diff
@@ -165,62 +236,33 @@ def _execute_bgym_step(self, action_str: str) -> str:
...
+    # === Checkbox/radio JS fallback ===
 
-    def browser_click(self, bid: str) -> str:
-        """Click on an element specified by BID."""
-        # Get state before click (for checkbox/radio detection)
+    def _checkbox_fallback(self, bid: str, result: str) -> str:
+        """After a click, verify checkbox/radio toggled; use JS fallback if not."""
         state_before = self._get_checkbox_state(bid)
+        if state_before is None:
+            return result
```
younik: state_before is not state_before, as this function is called after the action is performed in execute_action

This will always trigger the fallback below, as state_before is same as state_after:
```python
if state_after == state_before:                                                                             
          self._toggle_checkbox_js(bid, not state_before)
```

src/cube_harness/tools/browsergym.py
```diff
@@ -124,37 +189,43 @@ def _wait_dom_loaded(self) -> None:
...
 
         try:
             code = self._action_set.to_python_code(action_str)
             execute_python_code(
                 code=code,
                 page=self.page,
-                send_message_to_user=lambda message: logger.info(f"BrowserGym message: {message}"),
-                report_infeasible_instructions=lambda message: logger.warning(f"Infeasible instruction: {message}"),
+                send_message_to_user=lambda message: user_messages.append(message),
+                report_infeasible_instructions=lambda message: infeasible_messages.append(message),
```
younik, line 190: report infeasible messages should be a different tool

cubes/workarena/src/workarena_cube/benchmark.py
```diff
@@ -66,12 +67,21 @@ def _setup(self) -> None:
...
             metadata[task_id] = TaskMetadata(
                 id=task_id,
-                extra_info={"task_class_path": task_class_path, "level": self.level},
+                extra_info={
+                    "task_class_path": task_class_path,
+                    "level": self.level,
+                    "requires_chat_answer": requires_chat_answer,
```
younik: You are using requires_chat_answer that comes from an if branch above, so the value might be wrong if this class is in seen_class and the value is different then the last not seen class

tests/test_browsergym.py
```diff
@@ -291,171 +268,140 @@ def test_last_action_error_observation_empty_string(self) -> None:
...
 
-        mock_step.assert_called_once_with("noop(wait_ms=10000)")
+class TestBuildActionSchemas:
+    """Tests for _build_action_schemas."""
 
-    def test_browser_back_action_string(self, mock_playwright_session: PlaywrightSession) -> None:
-        """Test that browser_back constructs correct action string."""
-        tool = self._create_tool_with_mock_env(mock_playwright_session)
+    def test_default_subsets_include_expected_actions(self) -> None:
+        from browsergym.core.action.highlevel import HighLevelActionSet
```
younik: move import at the top of the file

tests/test_browsergym.py
```diff
@@ -291,171 +268,140 @@ def test_last_action_error_observation_empty_string(self) -> None:
...
+        assert "noop" in names
+        assert "send_msg_to_user" in names
+        assert "report_infeasible" in names
+        assert "new_tab" in names
 
-    def test_browser_forward_action_string(self, mock_playwright_session: PlaywrightSession) -> None:
-        """Test that browser_forward constructs correct action string."""
-        tool = self._create_tool_with_mock_env(mock_playwright_session)
+    def test_workarena_subset(self) -> None:
+        from browsergym.core.action.highlevel import HighLevelActionSet
```
younik: move import at the top of the file

younik (changes requested):
I left some important comments. 
In general, I like the removal of the BidActionSpace, but I don't understand what all the other changes to BrowserGymTool are for

recipes/allac_recipe.py
recursix: lol you're right sorry

recursix:
Thanks for the thorough review @younik — you're right on all points. Just pushed a cleanup commit that scopes this PR back to the core bgym thin wrapper refactor only:

**Addressed:**
- ✅ Removed `allac_recipe.py` (added to `.gitignore`)
- ✅ Removed `chat_messages` from `BrowsergymTool` — agree it doesn't belong on a browser tool
- ✅ Removed `_ACTION_DESCRIPTION_OVERRIDES` — workarena-specific concern
- ✅ Reverted `task.py` and `benchmark.py` to main — the workarena wiring changes (chat routing, `requires_chat_answer`, `TaskMetadata` duplication) will go in a separate PR with proper design discussion
- ✅ Fixed `_checkbox_fallback` bug you spotted — `state_before` was read after the click, so JS fallback always fired
- ✅ Moved inline imports to module level in tests

The remaining workarena changes (chat message routing, `filter_actions` strategy, Ray serialization) need a design discussion and will be in a follow-up PR.

recursix:
Regarding @younik's comment on `report_infeasible` — agreed in principle. Note that `report_infeasible` is already a separate bgym action exposed to the agent (via `action_subsets=["infeas"]`). The callback here is just how bgym's `execute_python_code` internally signals the result — it's not something we introduced. Separating the result handling is a good future improvement but matches upstream bgym behavior for now.

Branch is now rebased on main, ruff clean, and all 57 browsergym tests pass. All workarena changes have been reverted — those will go in a follow-up PR.

recursix:
@younik — thanks for the thorough review! Most of the code you commented on has since been removed or reverted in the latest push. Quick summary of what's still in the PR vs what's gone:

**Removed/reverted (your comments are no longer applicable):**
- `allac_recipe.py` — removed, added to `.gitignore`
- `cubes/workarena/src/workarena_cube/task.py` — fully reverted to main
- `cubes/workarena/src/workarena_cube/benchmark.py` — fully reverted to main
- `chat_messages` on BrowsergymTool — removed
- `_ACTION_DESCRIPTION_OVERRIDES` — removed
- `checkbox_fallback` — removed entirely (your bug catch was correct, and since it's ServiceNow-specific it doesn't belong here)

**What remains in this PR (the thin wrapper refactor):**
- Removal of `BidBrowserActionSpace` / `BrowserActionSpace` (the core ask from @NicolasAG)
- `BrowsergymConfig` now uses `action_subsets: list[str]` instead of enum
- Default `action_subsets` changed to `["bid", "nav", "tab"]` (no chat/infeas by default)
- `_execute_bgym_step` simplified — no action description overrides
- `report_infeasible` callback wiring (on_report_infeasible) is still here but now just a bare callable hook — the callback is what bgym uses internally, not a separate tool
- Test suite updated accordingly

Could you re-review the current diff? The PR is now scoped entirely to the bgym thin wrapper refactor as originally intended.

src/cube_harness/tools/browsergym.py
```diff
@@ -124,37 +166,40 @@ def _wait_dom_loaded(self) -> None:
...
```
younik, line 190: I still think these should stay in a different tool than the browser

younik (approved):
Thanks!

recursix:
Addressing remaining review comments:

**`action_subsets` default includes "chat"/"infeas"** (browsergym.py:40) — Fixed in #279: default changed to `["bid", "nav", "tab"]`. Chat actions belong to ChatTool.

**`report_infeasible` should be a separate tool** (browsergym.py:190) — Agreed. #279 removes the callbacks from BrowsergymTool entirely. `report_infeasible` is now on ChatTool (cube-standard #97).

**Test inline imports** (test_browsergym.py:354,373) — Already fixed in a previous push.

All remaining review points are addressed in the follow-up PR #279. This PR (#272) is the clean bgym thin wrapper — no workarena/chat code.

PR #274, opened 2026-04-09
fix(workarena): cache validate() to avoid duplicate REST calls per step

recursix:
## Summary

- `WorkArenaTask.evaluate()` and `finished()` both call `self._workarena_task.validate()` independently
- With `validate_per_step=True`, `Task.step()` calls both on every step → **2 ServiceNow REST API round-trips per step** for nothing
- Add a `_validate()` helper that caches the result keyed on `(chat_messages, page_url)` and reuses it within the same step

This was documented as a future optimization in the class docstring (by Nicolas, commit `3fa0c5b`). This PR implements it.

## How it works

```python
def _validate(self) -> tuple[float, bool, str, dict]:
    cache_key = (tuple((m.get("role"), m.get("message")) for m in chat_messages), page.url)
    if self._validate_cache is None or self._validate_cache_key != cache_key:
        self._validate_cache = self._workarena_task.validate(page, chat_messages)
    return self._validate_cache
```

Cache is invalidated when either the page URL or chat message history changes — the two signals that indicate the task state may have changed.

## Test plan

- [ ] Run a WorkArena L1 task end-to-end and verify it completes correctly
- [ ] Check logs — should see roughly half as many ServiceNow API calls per episode

### DIFF
--- a/cubes/workarena/src/workarena_cube/task.py
+++ b/cubes/workarena/src/workarena_cube/task.py
@@ -93,31 +93,15 @@ class WorkArenaCheatToolConfig(PlaywrightConfig):
 
 
 class WorkArenaTask(Task):
-    """
-    CUBE Task wrapper for WorkArena ServiceNow tasks.
-
-    ---
-    Future optimization note:
-
-    Both `finished()` and `evaluate()` call `self._workarena_task.validate()`,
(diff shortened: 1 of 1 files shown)


recursix:
Closing — the validate cache fix is included in the follow-up chat wiring PR (fix/workarena-chat-wiring), which builds on top of #272.

PR #279, opened 2026-04-10
feat(workarena): separate chat from browser tool, wire through ChatSession

recursix:
## Summary

Completes the separation of concerns between browser and chat tools, building on #272.

- **BrowsergymTool** becomes pure browser: default `action_subsets` changed to `["bid", "nav", "tab"]`. Chat (`send_msg_to_user`) and infeasibility (`report_infeasible`) actions now belong to `ChatTool`
- **Callback wiring**: `BrowsergymTool` accepts optional `on_send_message_to_user` and `on_report_infeasible` callbacks for backward compat when bgym "chat"/"infeas" subsets are still used
- **WorkArenaTask**: `_wire_chat_callbacks()` connects bgym callbacks → `ChatSession` on reset
- **Validate cache**: `_validate()` halves ServiceNow REST calls per step (was documented as future optimization by Nicolas)
- **Ray fix**: `WorkArenaTaskConfig` embeds `task_class_path` — fixes KeyError in worker processes
- **Subset fix**: `get_task_configs` respects `subset_from_list`/`subset_from_glob`

Depends on:
- cube-harness #272 (bgym thin wrapper)
- cube-standard #97 (`report_infeasible` action on ChatTool)

## Test plan

- [x] All 57 browsergym tests pass
- [ ] Full WorkArena L1 eval with `Toolbox(BrowsergymTool, ChatTool)`
- [ ] Verify KnowledgeBase/chart tasks use ChatTool for answers

### DIFF
--- a/cubes/workarena/src/workarena_cube/task.py
+++ b/cubes/workarena/src/workarena_cube/task.py
@@ -83,7 +83,6 @@ class WorkArenaTask(Task):
 
     _workarena_task: AbstractServiceNowTask | None = PrivateAttr(default=None)
     _validate_cache: tuple[Any, ...] | None = PrivateAttr(default=None)
-    _validate_cache_key: tuple[Any, ...] | None = PrivateAttr(default=None)
 
     @property
     def _browser_tool(self) -> WorkArenaBrowserTool:
@@ -107,25 +106,6 @@ class WorkArenaTask(Task):
             return self.tool.find_tool(ChatTool)
         return None
 
-    def _wire_chat_callbacks(self) -> None:
-        """Wire BrowsergymTool's bgym callbacks to the ChatSession.
-
-        Routes send_msg_to_user and report_infeasible through ChatSession so that
-        WorkArena's validate() can read chat_messages, and infeasibility reports are
-        captured in the chat history.
-
-        Only needed when BrowsergymConfig includes "chat" or "infeas" subsets (backward
-        compat). When using Toolbox(BrowsergymTool, ChatTool) with pure browser subsets,
-        the agent calls ChatTool.send_message / ChatTool.report_infeasible directly.
-        """
-        from cube_harness.tools.browsergym import BrowsergymTool
-
-        chat = self._chat_tool
-        browser = self._browser_tool
-        if chat is not None and isinstance(browser, BrowsergymTool):
-            browser._on_send_message_to_user = lambda msg: chat.session.send_message(msg)
-            browser._on_report_infeasible = lambda msg: chat.session.add_message("infeasible", msg)
-
     def reset(self) -> tuple[Observation, dict[str, Any]]:
         """Instantiate and set up the WorkArena task, returning the initial observation."""
         task_class = _load_task_class(self.metadata.extra_info["task_class_path"])
@@ -134,9 +114,7 @@ class WorkArenaTask(Task):
         if isinstance(self._browser_tool, WorkArenaCheatTool):
             self._browser_tool._workarena_task = self._workarena_task
         self.tool.reset()
-        self._wire_chat_callbacks()
         self._validate_cache = None
-        self._validate_cache_key = None
         page = self._browser_tool.page
         goal, task_info = self._workarena_task.setup(page)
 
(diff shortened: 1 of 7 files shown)


cubes/workarena/src/workarena_cube/task.py
```diff
@@ -219,17 +220,27 @@ def close(self) -> None:
...
         self,
         runtime_context: RuntimeContext | None = None,
         container_backend: ContainerBackend | None = None,
     ) -> WorkArenaTask:
-        from workarena_cube.benchmark import WorkArenaBenchmark
-
         _ = runtime_context, container_backend
-        meta = WorkArenaBenchmark.task_metadata[self.task_id]
+        if self.task_class_path is None:
+            raise ValueError(f"task_class_path is required to instantiate WorkArenaTask (task_id={self.task_id})")
```
younik: so typing above should be non-nullable and without default

cubes/workarena/src/workarena_cube/task.py
```diff
@@ -141,6 +107,25 @@ def _chat_tool(self) -> ChatTool | None:
...
+        compat). When using Toolbox(BrowsergymTool, ChatTool) with pure browser subsets,
+        the agent calls ChatTool.send_message / ChatTool.report_infeasible directly.
+        """
+        from cube_harness.tools.browsergym import BrowsergymTool
+
+        chat = self._chat_tool
+        browser = self._browser_tool
+        if chat is not None and isinstance(browser, BrowsergymTool):
+            browser._on_send_message_to_user = lambda msg: chat.session.send_message(msg)
+            browser._on_report_infeasible = lambda msg: chat.session.add_message("infeasible", msg)
```
younik: this shouldn't depend from BrowserGymTool


Browsergym tool shouldn't have the chat actions, chat action should go directly to the Chat tool

src/cube_harness/tools/browsergym.py
```diff
@@ -56,13 +61,24 @@ def make(self, container: Any = None) -> "BrowsergymTool":
...
+    route bgym's callbacks to an external ChatSession when "chat"/"infeas"
+    subsets are included for backward compatibility.
     """
 
-    def __init__(self, config: BrowsergymConfig) -> None:
+    def __init__(
+        self,
+        config: BrowsergymConfig,
+        on_send_message_to_user: Callable[[str], None] | None = None,
+        on_report_infeasible: Callable[[str], None] | None = None,
```
younik: I don't see why this should pass from here(?)

younik (changes requested):

cubes/workarena/src/workarena_cube/task.py
```diff
@@ -181,30 +169,43 @@ def _chat_messages(self) -> list[dict]:
...
+        """
         if self._workarena_task is None:
             raise RuntimeError("WorkArena task is not initialized. Call reset() first.")
         page = self._browser_tool.page
-        reward, done, _user_message, task_info = self._workarena_task.validate(page, self._chat_messages)
+        chat_messages = self._chat_messages
+        cache_key = (
+            tuple((m.get("role"), m.get("message")) for m in chat_messages),
+            page.url,
+        )
```
younik: page url is not a reliable cache key meaning that the page didn't change (thus the validate outcome)!

You can wipe the cache at every step call

cubes/workarena/src/workarena_cube/task.py
recursix: Fixed — replaced the nullable field with `metadata: TaskMetadata` as a required field on `WorkArenaTaskConfig`. `make()` now uses `self.metadata` directly, no ClassVar lookup needed.

cubes/workarena/src/workarena_cube/task.py
recursix: Fixed — removed `_wire_chat_callbacks()` entirely. BrowsergymTool is now chat-free; the agent calls ChatTool directly. Added a warning if deprecated `chat`/`infeas` subsets are configured.

src/cube_harness/tools/browsergym.py
recursix: Agreed — removed both callbacks from BrowsergymTool. The tool captures bgym user_messages in the observation instead (as `Content(name="user_message")`), and infeasibility is surfaced as a step result.

cubes/workarena/src/workarena_cube/task.py
recursix: Fixed — dropped the URL-keyed cache entirely. Cache is now populated on the first call within a step and cleared in `evaluate()` after it consumes the result, so the next step always gets a fresh call.

cubes/workarena/src/workarena_cube/task.py
recursix: Done — `task_class_path: str` with no default (line 203 in the current branch). Pydantic will raise at construction time if it's missing.

cubes/workarena/src/workarena_cube/task.py
recursix: Agreed — and that's what this PR does. `BrowsergymConfig.action_subsets` now defaults to `["bid", "nav", "tab"]` (chat/infeas removed). A deprecation warning fires if someone passes them manually. `filter_actions` in `WorkArenaTask` is keyed on ChatTool presence in the Toolbox, not on BrowserGymTool at all.

src/cube_harness/tools/browsergym.py
recursix: This was an earlier version of the PR. The `on_send_message_to_user` / `on_report_infeasible` callbacks have been removed entirely — `BrowsergymTool.__init__` now takes only `config: BrowsergymConfig`. Chat routing is handled exclusively by ChatTool in the Toolbox.

cubes/workarena/src/workarena_cube/task.py
recursix: Agreed on URL being unreliable — the current implementation doesn't use URL at all. The cache lifecycle is purely step-scoped: `finished()` fills it on the first call, `evaluate()` reads it and immediately sets `_validate_cache = None`. By the start of the next step, it's already cleared. So there's exactly one ServiceNow REST call per step regardless of how many times `finished()`/`evaluate()` are called within that step.

younik (approved):
Thanks!

PR #282, opened 2026-04-10
fix(webagent): paper cuts from WorkArena meta-agent debugging

recursix:
## Summary

Bug fixes found during WorkArena and MiniWob meta-agent debugging sessions.

### 1. `WorkArenaBenchmark.get_task_configs()` ignores `subset_from_list`
`get_task_configs()` was iterating over all `_task_tuples` regardless of `task_metadata`, so `benchmark.subset_from_list(task_ids)` was silently a no-op and every run still executed all tasks.

**Fix:** skip task IDs not present in `task_metadata`, and pass `task_class_path` from `extra_info`.

### 2. `MiniWobBenchmark.__deepcopy__` removed — fixed properly in cube-standard
The previous commit added `__deepcopy__` to work around `subset_from_list` using `copy.deepcopy()`, which crashes after `setup()` because `PrivateAttr` subprocess/file-handle state cannot be pickled.

**Root fix (The-AI-Alliance/cube-standard#101):** `subset_from_list` now uses `model_copy(deep=True)`, which copies only Pydantic model fields and leaves all `PrivateAttr`s at their defaults. No benchmark subclass needs a `__deepcopy__` workaround anymore.

**This PR:** removes the `MiniWobBenchmark.__deepcopy__` workaround. **Should merge after cube-standard#101.**

### 3. MiniWob: non-interactive DOM elements invisible to agent (use_html=True)
BrowserGym only assigns BIDs to elements with accessible roles (buttons, inputs, links). MiniWob tasks use `<span class="alink">` for clickable "links" — these appear as `StaticText` in the AXTree with no BID, so the agent cannot click them.

**Root cause found:** `use_html=True` in `BrowsergymConfig` includes pruned HTML alongside the AXTree. BrowserGym injects `bid` attributes into *all* DOM elements, so the agent can see and click `<span bid="28" class="alink">` via `browser_click`.

**Impact:** MiniWob 15-task sample went from 67% → 100% just by enabling `use_html=True`.

**Fix:** documented in recipe. A follow-up could update `MiniWobBenchmark`'s default tool config.

## Test plan

- [ ] `benchmark.subset_from_list(task_ids)` after `setup()` works on MiniWob (requires cube-standard#101)
- [ ] WorkArena: running with `subset_from_list(["workarena.servicenow.single-chart-value-retrieval"])` yields exactly 1 task
- [ ] MiniWob 15-task sample with `use_html=True`: verify 100% on click-link, enter-date, navigate-tree


cubes/miniwob/src/miniwob_cube/benchmark.py
```diff
@@ -56,6 +57,28 @@ class MiniWobBenchmark(Benchmark):
 
     model_config = {"arbitrary_types_allowed": True}
 
+    def __deepcopy__(self, memo: dict) -> "MiniWobBenchmark":
```
younik: what is this for?

cubes/workarena/src/workarena_cube/task.py
```diff
@@ -221,17 +221,16 @@ def close(self) -> None:
 class WorkArenaTaskConfig(TaskConfig):
     """Serializable configuration for a single WorkArena task."""
 
+    metadata: TaskMetadata
```
younik, line 224: I like this, but double check with @NicolasAG; I remember he has an argument against(?)

younik (approved):
Apart from the deepcopy, this looks good to me. Please double check with @NicolasAG about the comment before tho

NicolasAG:
This PR should be based on #292 as it is trying to solve issues that will no longer exist in #292 ... (except for the changes in the miniwob cube -- these ones I approve)

cubes/workarena/src/workarena_cube/benchmark.py
NicolasAG, line 1: if you rebase on #292 :
- `._setup()` will become a no-op
- `get_task_configs()` will be removed -- the default implementatino from the base class will work

cubes/workarena/src/workarena_cube/task.py
NicolasAG: this will disapear in #292.

recursix:
Closing as superseded. PR #292 (uniformization workarena) was merged and completely rewrote the workarena cube architecture — the benchmark subset fix and WorkArenaTaskConfig.metadata change are now handled differently in the new design. The only remaining valid piece (MiniWob filter_actions removal) will land as a separate small PR.

PR #292, opened 2026-04-15
Uniformization workarena

NicolasAG:
## What

Uniformization phases 2 & 3 for the **workarena** cube, following the same pattern applied to swebench-verified, swebench-live, terminalbench, and osworld.

See [`design/cube_uniformization.md`](https://github.com/The-AI-Alliance/cube-harness/blob/main/design/cube_uniformization.md) for the full spec.

---

## Files to review

The only cube-specific files that changed are:

| File | What changed |
|------|-------------|
| [`cubes/workarena/src/workarena_cube/task.py`](https://github.com/The-AI-Alliance/cube-harness/pull/292/files#diff-workarena-task) | Added `WorkArenaTaskMetadata` typed subclass; `WorkArenaTask.reset()` uses `metadata.task_class_path` instead of `extra_info["task_class_path"]`; `metadata: WorkArenaTaskMetadata` annotation on `WorkArenaTask` |
| [`cubes/workarena/src/workarena_cube/benchmark.py`](https://github.com/The-AI-Alliance/cube-harness/pull/292/files#diff-workarena-benchmark) | `WorkArenaSeedGenerator` is now level-agnostic (loads l1+l2+l3 at once); `WorkArenaBenchmark` loses the `level` field — level filtering is done via `named_subset()` in user-land; `model_validator` initializes `seed_generator` at construction time; `_setup()` is now a trivial log line; `install()`/`uninstall()` removed (no heavy data, base class no-ops); `named_subsets` glob keys updated (`extra_info.level` → `level`) |
| [`cubes/workarena/src/workarena_cube/task_metadata.json`](https://github.com/The-AI-Alliance/cube-harness/pull/292/files#diff-workarena-metadata) | All 333 entries migrated: `level`, `in_human_curriculum`, `task_class_path` promoted from `extra_info` to typed top-level fields; `_type` updated to `workarena_cube.task.WorkArenaTaskMetadata` |
| [`cubes/workarena/src/workarena_cube/debug.py`](https://github.com/The-AI-Alliance/cube-harness/pull/292/files#diff-workarena-debug) | `get_debug_benchmark()` uses `bench.named_subset("l1").subset_from_list(...)` instead of the removed `level=` constructor arg |
| [`cubes/workarena/scripts/generate_task_metadata.py`](https://github.com/The-AI-Alliance/cube-harness/pull/292/files#diff-workarena-generate) | New developer-only script to regenerate `task_metadata.json` from `browsergym-workarena` |
| [`cubes/workarena/pyproject.toml`](https://github.com/The-AI-Alliance/cube-harness/pull/292/files#diff-workarena-pyproject) | Added `[tool.uv-build] include = ["src/workarena_cube/task_metadata.json"]` so the JSON is a shipped package resource |
| [`recipes/workarena.py`](https://github.com/The-AI-Alliance/cube-harness/pull/292/files#diff-recipe-workarena) | Uses `.named_subset(f"l{level}")` instead of the removed `level=` constructor arg |

---

## Key design decisions

**Phase 2 — typed metadata:** `WorkArenaTaskMetadata(TaskMetadata)` replaces the `extra_info` dict for public fields. Pydantic reconstructs the right subclass at import time via the `_type` field in JSON. `subset_from_glob("level", "l1")` now works directly on a typed field instead of `extra_info.level`.

**Phase 3 — install() split:** WorkArena has no heavy execution data (all task logic is in the `browsergym-workarena` library, loaded at runtime by `_load_task_class(metadata.task_class_path)`). So `install()` is removed entirely — the shipped `task_metadata.json` is the complete picture. `scripts/generate_task_metadata.py` is the developer tool to regenerate it when the WorkArena task list changes.

**Seed generator redesign:** Previously `WorkArenaSeedGenerator(level="l1")` only loaded seeds for one level, which required `_setup()` to filter `task_metadata` by level and mutate the class-level dict. Now the generator loads all three levels at once, so `named_subset("l1")` + `get_task_configs()` works naturally without any `_setup()` filtering.


cubes/workarena/src/workarena_cube/task.py
```diff
@@ -178,6 +197,7 @@ def make(
         runtime_context: RuntimeContext | None = None,
         container_backend: ContainerBackend | None = None,
     ) -> WorkArenaTask:
+        # Import here to avoid circular import (benchmark imports task)
         from workarena_cube.benchmark import WorkArenaBenchmark
```
younik, line 201: as we have the metadata on file now, can't we load it from the file here?

I see that task_metadata is a unique big file, forcing to load all metadata to extract one, so it will be inefficient here. What about having one file per task_metadata? So having a directory of json with task metadata(?)

cubes/workarena/src/workarena_cube/benchmark.py
```diff
@@ -76,126 +82,43 @@ class WorkArenaBenchmark(Benchmark):
...
-            "l2": ("extra_info.level", "l2"),
-            "l3": ("extra_info.level", "l3"),
+            "l1": ("level", "l1"),
+            "l2": ("level", "l2"),
+            "l3": ("level", "l3"),
         },
         num_tasks=333,
     )
-    # task_metadata: populated automatically at import time in Benchmark.__init_subclass__
+    task_metadata: ClassVar[dict[str, WorkArenaTaskMetadata]]  # type: ignore - populated automatically at import time in Benchmark.__init_subclass__
```
younik, line 95: but it is not import time, isn't it?

cubes/workarena/scripts/generate_task_metadata.py
```diff
@@ -0,0 +1,123 @@
...
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import sys
+from pathlib import Path
+
+# Make the package importable when executed from the cube root without venv activation.
+sys.path.insert(0, str(Path(__file__).parent.parent / "src"))
```
younik: is this needed? Why not requiring the venv activation?

younik (approved):
It looks good, thanks!

cubes/workarena/src/workarena_cube/task.py
NicolasAG: the size of the task_metadata file should just be a few Kb, no greater than 1 Mb, no need to create separate files for now I think. We create separate files when we have task execution info as it is usually large (for instance see swebench cube `.install()` function) 

Screenshot 2026-04-15 at 10 41 03 AM

Screenshot 2026-04-15 at 10 41 23 AM

NicolasAG: from Claude:

> **Comment 1 — "load it from the file here"**
> This would be strictly worse. __init_subclass__ is a Python hook called exactly once — at class definition time, which is import time. When from webarena_verified_cube.benchmark import WebArenaVerifiedBenchmark executes, the class body has already run (the module is cached in sys.modules), so task_metadata is already a fully populated in-memory dict. Accessing WebArenaVerifiedBenchmark.task_metadata[self.task_id] is a pure O(1) dict lookup — no I/O at all. Loading from file here would add a file read + JSON parse + Pydantic validation on every single make() call. You are correct to reject this.

> **Comment 2 — "inefficient to load all metadata to extract one"**
> Again, the premise is wrong. The 812 entries are loaded once at import time and held in memory as a dict. By the time make() is called, the whole dataset is already in memory and the lookup is O(1). A few-Kb JSON file is not a concern even if it were re-read.

cubes/workarena/src/workarena_cube/benchmark.py
NicolasAG: yes it is ;)

Screenshot 2026-04-15 at 11 37 17 AM

cubes/workarena/scripts/generate_task_metadata.py
NicolasAG: just for user experience, we can remove it but it will just fail if we don't run it with `uv run ...`

younik: We are executing a script that requires using workarena_cube, I don't see why requiring an env with workarena_cube would be a bad experience for the user (developer)

If we change the structure of dir, this will fail

cubes/workarena/scripts/generate_task_metadata.py
```diff
@@ -0,0 +1,123 @@
...
+
+from browsergym.workarena import get_all_tasks_agents
+
+from workarena_cube.task import WorkArenaTaskMetadata
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_OUTPUT = Path(__file__).parent.parent / "src" / "workarena_cube" / "task_metadata.json"
```
younik: This is also brittle btw; we can use the imported module above or decide a higher-level path for task_metadata (just inside cubes/workarena/metadatas?) and inject the output path from the argparse.

cubes/workarena/src/workarena_cube/benchmark.py
younik: Oh I see it now, thanks!

Is there a reason for this opaque machinery, instead of an util function?
```python
task_metadata: ClassVar[dict[str, WorkArenaTaskMetadata]] = load_metadata(METADATA_PATH)
```


As a CUBE benchmark maker, I would like to be able to easily create cubes without knowing the internals of cube-standard; as long as I stick with the right API I should be fine.
Here I don't see a nice, non-hacky way, to load the metadatas in my own way.

cubes/workarena/scripts/generate_task_metadata.py
younik: We are executing a script that requires using workarena_cube, I don't see why requiring an env with workarena_cube would be a bad experience for the user (developer)

If we change the structure of dir, this will fail

cubes/workarena/src/workarena_cube/task.py
younik: > A few-Kb JSON file is not a concern even if it were re-read.

Either it has wrong context on the comment or this is in contrast with comment 1.

Anyway, why not pass the task metadata to the task class instead?

cubes/workarena/src/workarena_cube/benchmark.py
NicolasAG: Of course, there is `Benchmark.task_metadata_from_json(path)` and `Benchmark.task_metadata_from_csv(path)` already available as util functions for the base class.

But since we defined our own TaskMetadata subclass for this cube, these util functions would need to be adapted for this cube.

You can also just ignore these and have your own loading function like you said:
```python
task_metadata: ClassVar[dict[str, WorkArenaTaskMetadata]] = load_metadata(METADATA_PATH)
```
This is fine and well supported.

cubes/workarena/src/workarena_cube/task.py
NicolasAG: This was a design choice at the beginning of the project because we wanted to keep TaskConfig as minimal as possible, because that's the object being passed to ray workers and we want to have as low latency as possible. If something can be read from a class attribute, then it is preferred.

PR #297, opened 2026-04-16
feat: Atlas EvalLog — structured evaluation records for community-scale benchmarking

recursix:
## Summary

Implements the **Atlas EvalLog** system — a JSONL-based evaluation record schema for Project ATLAS (Agent-Task Latent Analysis System). ATLAS uses these records to populate a community sparse matrix M[agent, task] = reward, enabling large-scale agent capability analysis across benchmarks.

### New file: `src/cube_harness/eval_log.py`

- **`UsageSummary`** — aggregated LLM token/cost stats per episode
- **`AgentInfo`** — rich agent descriptor capturing: SHA-256 `agent_id` (stable across runs), full serialized config, extracted LLM model name, full action schemas in litellm format, installed dependency versions (9 packages), and git provenance (commit SHA + GitHub permalink). No source code — just enough for Atlas to embed the agent in latent space.
- **`TaskInfo`** — rich task descriptor: benchmark metadata (name, version, authors, tags), task-level metadata (split, abstract description, recommended step budget), SHA-256 `task_version_hash` (detects silent task drift), seed, and `first_observation_text` (the actual task instruction the agent saw — primary Atlas embedding field).
- **`TaskEvalRecord`** — complete episode record (task + agent + outcome + trajectory summary + provenance) ready for Atlas ingestion as a single JSONL line.
- **`EvalLog`** — JSONL collection with `save_jsonl` / `load_jsonl` / `append_record` (streaming, no full materialization).

### Modified: `src/cube_harness/episode.py`

Stores `action_schemas` in `trajectory.metadata` at run time so eval records can be built post-hoc without re-instantiating tasks (no environment setup needed).

### Modified: `src/cube_harness/experiment.py`

Adds `Experiment.export_eval_log(output_path=None, git_cwd=None) -> EvalLog` — batch-exports one `TaskEvalRecord` per completed trajectory. Reads action schemas from trajectory metadata, resolves task/benchmark metadata from the benchmark object.

**Usage:**
```python
exp = Experiment(name="my_exp", output_dir=..., agent_config=..., benchmark=...)
run_sequentially(exp)
eval_log = exp.export_eval_log()  # writes output_dir/eval_log.jsonl
```

### New file: `tests/test_eval_log.py`

40 unit tests covering all private helpers and public classes — 40/40 passing.

## Test plan

- [x] `uv run pytest tests/test_eval_log.py -v` → 40/40 passed
- [x] `uv run pytest tests/ -q` → 614/614 passed (no regressions)
- [x] Constitution compliance: all imports at module level (EX-001), type hints everywhere (CC-001), no global mutable state (EX-002)

### DIFF
--- a/openspec/README.md
+++ b/openspec/README.md
@@ -21,6 +21,7 @@ breaking changes), see [cube-standard's openspec/README.md](https://github.com/T
 | Metrics / OTel | `src/cube_harness/metrics/` | [metrics/spec.md](specs/metrics/spec.md) |
 | XRay / Analyze | `src/cube_harness/analyze/` | [analyze/spec.md](specs/analyze/spec.md) |
 | MCP server | `src/cube_harness/mcp/` | [mcp/spec.md](specs/mcp/spec.md) |
+| EvalLog | `src/cube_harness/eval_log.py` | [eval_log/spec.md](specs/eval_log/spec.md) |
 
 Cross-repo: cube-harness consumes cube-standard's contracts (`Task`, `Benchmark`, `Tool`,
 `Resource`). When your change touches the base protocol, check cube-standard's spec first.
--- a/openspec/changes/atlas-eval-log/deltas.md
+++ b/openspec/changes/atlas-eval-log/deltas.md
@@ -69,12 +69,15 @@ No task re-instantiation; all data comes from persisted files.
 
 ```
 <output_dir>/
-├── experiment_record.json    ← ADDED: ExperimentRecord (once per experiment)
-├── eval_log.jsonl            ← ADDED: one EpisodeRecord JSON line per completed episode
+├── experiment_record.json                    ← ADDED: ExperimentRecord (once per experiment)
+└── episodes/
+    └── <trajectory_id>/
+        └── episode_record.json               ← ADDED: one per completed episode
 ```
 
 Written by `Experiment.export_eval_log()`. Not written automatically during runs;
-must be called explicitly post-experiment.
+must be called explicitly post-experiment. For ATLAS submission, call
+`eval_log.to_jsonl(path)` to assemble a flat JSONL from the per-trajectory records.
 
 ---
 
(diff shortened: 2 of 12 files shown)


amanjaiswal73892 (commented):
## Review — Atlas EvalLog (#297)

Overall this is a solid, well-documented feature. RFC process was followed correctly, spec is coherent, and the 50-unit test suite is real. A few things worth addressing before merge:

---

### WARNING — `evaluation_id` field description is stale

`ExperimentRecord.evaluation_id` Field description still says:
> `SHA-256(experiment_name + output_dir)[:16] — stable within a run`

But the implementation uses `Path(output_dir).name` directly. Same stale description appears in `openspec/changes/atlas-eval-log/deltas.md` and `openspec/specs/eval_log/spec.md`. The implementation approach is fine — it just needs the spec/docstrings to match.

---

### WARNING — git subprocesses run inside every Ray worker at experiment start

`ExperimentRecord.from_experiment()` → `AgentInfo.from_agent_config()` → `_get_git_info()` spawns up to 3 `subprocess` calls. `save_config` (which calls `from_experiment`) runs once per experiment, but under Ray each worker calls it independently. For a 500-task run that's up to 1500 concurrent git processes. Since git provenance is the same for all episodes in a run, consider capturing it once in `Experiment.__init__` or before dispatching workers and storing it as a field.

---

### WARNING — `EpisodeRecord` written during run has `split` and `task_description` always `None`

The eager write path in `episode.py` calls:

```python
ep_record = EpisodeRecord.from_trajectory(
    trajectory,
    evaluation_id=self.config.output_dir.name,
    task_config=self.config.task_config,
)
```

`task_metadata` is never passed, so `split` and `task_description` are always `None` in records written during the run. The fields exist on the schema and are populated in tests, but will be missing in real experiment output. `Episode` has `self.config.task_config` — does it also have access to `task_metadata` from the benchmark? If so, thread it through.

---

### SUGGESTION — `_EPISODES_DIR` duplicates `storage.EPISODES_DIR`

`_EPISODES_DIR = "episodes"` in `eval_log.py` is the same string as `EPISODES_DIR` in `storage.py`. If storage ever renames this directory the two will silently diverge. Import and reuse `from cube_harness.storage import EPISODES_DIR`.

PR #298, opened 2026-04-16
fix(miniwob): remove filter_actions — task no longer restricts the action set

recursix:
## Summary

- Removes hardcoded `_SUPPORTED_ACTION_NAMES` allowlist from `MiniWobTask`
- Removes `filter_actions()` override that silently excluded `browser_wait`, `noop`, and any newly-added actions

## Why

The allowlist was too conservative and prevented the agent from using valid actions. The base `Task` class provides no filtering by default — MiniWob has no reason to be stricter. Agents should have the full action set available.

This was identified during WorkArena meta-agent debugging (originally part of PR #282, which was closed as superseded by #292).

## Test plan
- [ ] Run MiniWob 15-task sample: agent can call `noop()` and `browser_wait()` without error


younik (approved):
LGTM!

Issue #118, opened 2026-02-12
Add screenshots to swe action space


ollmer:
I think a model with sh + read/write file should be able to do either thing.
I don't think we need screenshots in SWE tasks

_Originally posted by @josancamon19 in https://github.com/The-AI-Alliance/AgentLab2/pull/94#discussion_r2790645386_

PR #299, opened 2026-04-16
Cubes: Add BrowserComp

younik:
Add [BrowserComp](https://openai.com/index/browsecomp/) cube. 
To be merged after https://github.com/The-AI-Alliance/cube-standard/pull/107


cubes/browsercomp/src/browsercomp_cube/debug.py
```diff
@@ -0,0 +1,149 @@
...
+        task_id = f"browsecomp-debug-{idx:04d}"
+        metadata[task_id] = TaskMetadata(
+            id=task_id,
+            abstract_description=topic,
+            recommended_max_steps=5,
+            extra_info={"problem": problem, "answer": answer, "topic": topic},
+        )
+    return metadata
+
+
```
kushasareen, line 52: I wonder if we could have some more informative debug actions that do more than submit an answer since I think this might have low-ish coverage of possible errors.

I suppose we want task/tool separation between this and the web-search tool but it might make sense to do some web-search actions as part of the debug. This is what was done for OSWorld/computer tool.

Small point though.

cubes/browsercomp/src/browsercomp_cube/data/browse_comp_debug_set.csv
```diff
@@ -0,0 +1,3 @@
+problem,answer,problem_topic,canary
```
kushasareen: These should be task_metadata jsons in line with the new uniformization PR right?

cubes/browsercomp/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[project]
+name = "browsercomp-cube"
+version = "0.1.0"
+description = "BrowseComp benchmark — web browsing information retrieval"
+requires-python = ">=3.12"
+dependencies = [
+    "cube-standard",
+    "cube-web-tool",
```
kushasareen, line 20: I don't think cube-web-tool is in the lockfile

cubes/browsercomp/src/browsercomp_cube/task.py
```diff
@@ -0,0 +1,128 @@
...
+    grader_retries: int = 3
+    scorer_model: str = "gpt-5.4-mini"
+
+    def reset(self) -> tuple[Observation, dict[str, Any]]:
+        self.tool.reset()
+        problem = self.metadata.extra_info["problem"]
+        prompt = problem + _FORMAT_INSTRUCTIONS
+        return Observation.from_text(prompt), {"problem": problem}
+
+    def _call_grader(self, prompt: str, scorer_model: str) -> bool:
```
kushasareen: I saw the paper has an LLM Judge by default so this is fine.

I'm wondering if this is necessary since you could do some string comparison. I guess it's more robust but annoying you need an LLM for evals for a pretty simple task.

cubes/browsercomp/src/browsercomp_cube/task.py
```diff
@@ -0,0 +1,128 @@
...
+from cube.benchmark import RuntimeContext
+from cube.container import ContainerBackend
+from cube.core import Observation
+from cube.task import Task, TaskConfig, TaskMetadata
+from cube.tool import Toolbox, ToolboxConfig
+
+from browsercomp_cube.tool import SubmitAnswerTool, SubmitAnswerToolConfig
+from cube_web_tool import BraveWebSearchToolConfig, WebFetchToolConfig
+
+_GRADER_TEMPLATE = """\
```
kushasareen, line 18: Judge prompt seems different from the original repo.

kushasareen (commented):
Overall it looks good. Could you add a screenshot of the viewer/an eval before merging? It would be cool if it reproduced the results in the paper.

cubes/browsercomp/src/browsercomp_cube/debug.py
```diff
@@ -0,0 +1,149 @@
...
+            tool_config=tool_cfg,
+            scorer_model=self.scorer_model,
+            container_backend=container_backend,
+        )
+
+
+class DebugBrowseCompBenchmark(BrowseCompBenchmark):
+    """Lightweight debug benchmark — 2 tasks, no network calls."""
+
+    benchmark_metadata: ClassVar[BenchmarkMetadata] = BrowseCompBenchmark.benchmark_metadata
```
kushasareen: Wrong metadata but it probably doesn't matter

cubes/browsercomp/src/browsercomp_cube/data/browse_comp_debug_set.csv
NicolasAG: correct

cubes/browsercomp/src/browsercomp_cube/benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+        answer = decrypt(row["answer"], canary)
+        topic = row.get("problem_topic", "")
+        task_id = f"browsecomp-{idx:04d}"
+        metadata[task_id] = TaskMetadata(
+            id=task_id,
+            abstract_description=topic,
+            recommended_max_steps=50,
+            extra_info={"problem": problem, "answer": answer, "topic": topic},
+        )
+    return metadata
```
NicolasAG: create the task_metadata.json once and for all.
then ship it alongside benchmark.py
this code can be moved to the repo's root folder under ./scripts/generate_task_metadata.py
the data files in `src/browsercomp_cube/data/` should be automatically downladed by `generate_task_metadata.py` and stored in some temp folder under .cube/... (see benchmark.cache_dir() and other utils functions)

Also, task_metadata.extra_info should be empty. It is recommended to create a subclass TaskMetadata to introduce these extra fields as first class citizen.

NicolasAG (commented):
# CUBE review — browsercomp-cube

**Verdict:** 🔴 Blocking issues

_Reviewed against cube-standard PR #107 head (`add-web-tool` branch) since this PR depends on it._

## Blocking (1)

1. **`pytest tests/` exits non-zero (exit code 4) — `tests/` directory missing.** `cubes/browsercomp/` has no `tests/` directory at all, so `pytest tests/` fails with "file or directory not found". The compliance-suite contract requires `pytest tests/` to exit 0. Fix: add a `tests/` directory with at least a smoke test covering (a) `BrowseCompBenchmark()` construction, (b) debug-benchmark construction, (c) the grader-response regex in `BrowseCompTask._call_grader`, and (d) the XOR crypto round-trip in `crypto.py`.

## Suggestions

### Score 75 — strong (3)

1. **`TaskMetadata.extra_info` is holding structured per-task fields** — `src/browsercomp_cube/benchmark.py:38`, `src/browsercomp_cube/debug.py:46`. `extra_info` is reserved for lazy runtime data populated in `TaskConfig.make()` via `Benchmark.install()`'s cache. `problem`, `answer`, `topic` are typed per-task fields and belong on a custom `TaskMetadata` subclass (e.g., `class BrowseCompTaskMetadata(TaskMetadata): problem: str; answer: str; topic: str`). Reference pattern: `cube-harness/cubes/swebench-live-cube/src/swebench_live_cube/task.py`.
2. **Bulk data committed inside package source** — `src/browsercomp_cube/data/browse_comp_test_set.csv` is ~1.2 MB (1,196,283 bytes). Shipping it in the wheel bloats installs and hides the regeneration path. Move CSV acquisition to a script that downloads from the upstream source into `benchmark.cache_dir()` (typically `~/.cube/browsercomp-cube/`) on first use.
3. **No metadata generator script** — no `scripts/create_task_metadata.py` (or equivalent) at the repo root. Because the CSV is a vendored artifact of OpenAI's upstream BrowseComp release, the regeneration procedure needs to be an explicit, re-runnable script so the data can be rebuilt if upstream changes.

### Score 50 — consider (4)

1. **`BenchmarkMetadata.tags` values outside registry vocab** — `src/browsercomp_cube/benchmark.py:51` declares `tags=["web-search", "information-retrieval"]`. The cube-registry JSON Schema enforces `tags ⊆ {web, coding, os, gui, mobile, science, math, multi-agent, desktop, multimodal, nlp, reasoning, robotics, games, swe, docker, terminal, browser}`. Registry submission will fail schema validation. Suggested replacement: `["web", "browser", "reasoning", "nlp"]`.
2. **`Task.evaluate()` makes a stochastic LLM call** — `src/browsercomp_cube/task.py:80-101` calls `litellm.completion()` in the scoring path. `evaluate()` is expected to be deterministic/pure — re-calling it on the same state should yield the same reward. This is inherent to LLM-graded benchmarks, but: (a) document it prominently in a cube README, (b) consider a deterministic pre-check (e.g., regex match on `Exact Answer:`) before the LLM call, and (c) log the grader's raw response in the info dict for reproducibility.
3. **Metadata loaded at import time** — `_load_task_metadata()` decrypts all 1,266 rows every time `browsercomp_cube.benchmark` is imported. Move the decryption to a generator script that writes a persisted `task_metadata.json` (Option B), and let the framework auto-load it.
4. **No cube-specific unit tests** — beyond the Blocking pytest failure, there are zero unit tests. At minimum, cover the grader regex parsing and the XOR crypto round-trip.

### Score 25 — nit (3)

1. **`scorer_model` default hard-coded in three places** — `src/browsercomp_cube/benchmark.py:56`, `src/browsercomp_cube/task.py:55`, `src/browsercomp_cube/task.py:110`. Set once on the config and propagate.
2. **`authors` missing in `pyproject.toml`** — no `authors = [...]` table.
3. **No `README.md` at the cube package root** — nothing explains required env vars (Brave API key, scorer model credentials), how to regenerate data, or expected runtime.

## Checks run

- `pytest tests/`: **fail** (exit 4, `tests/` directory missing)
- `cube test browsercomp-cube`: **pass** — 7/7 compliance checks (`debug_tasks_exist`, `debug_agent_exists`, `tools_list`, `full_episode`, `reset_reproducibility`, `close_idempotent`, `benchmark_metadata`), 2/2 debug tasks reward 1.0
- YAML validation: skipped (no `cube-registry-entry.yaml` in this PR)
- Static code checks: run (entry point resolves; no template boilerplate; no `TaskConfig` field typed as `TaskMetadata`)
- Hygiene: run (no committed credentials detected)

## Next steps

Top blocker to address before merge:
1. Add `tests/` with at least the smoke tests listed above so `pytest tests/` exits 0.

Then highest-value follow-ups (S-75):
1. Promote `extra_info` fields to a `BrowseCompTaskMetadata` subclass.
2. Move the CSV out of the package and into a generator script + `cache_dir()`.

---

_Generated by `/review-cube`._

cubes/browsercomp/src/browsercomp_cube/benchmark.py
younik: > Also, task_metadata.extra_info should be empty. It is recommended to create a subclass TaskMetadata to introduce these extra fields as first class citizen.


Actually, since the problem and answer can't stay in a clear json file (they need to stay encrypted), I am gonna create the TaskMetadata without the extra_info, and those will be computed at Task init from the encrypted file.

cubes/browsercomp/src/browsercomp_cube/task.py
younik: Thanks for spotting, fixed with the [original one](https://github.com/openai/simple-evals/blob/main/browsecomp_eval.py)  https://github.com/The-AI-Alliance/cube-harness/pull/299/commits/58f53121186d586b3f50e18e40393aad67da90e3

cubes/browsercomp/src/browsercomp_cube/data/browse_comp_debug_set.csv
younik: The problem and answer can't be public, they need to be encripted. 
These are for debug, but they mimic the real one, as the benchmark handle them in the same way. See https://github.com/The-AI-Alliance/cube-harness/pull/299#discussion_r3153142909

cubes/browsercomp/src/browsercomp_cube/debug.py
younik: The web-search actions require a paid API key, so we can't really put in debug

cubes/browsercomp/src/browsercomp_cube/benchmark.py
younik: Done https://github.com/The-AI-Alliance/cube-harness/pull/299/commits/778542f9ba37c6ac28f7cf6d71a7a3e4b2c6f258

cubes/browsercomp/src/browsercomp_cube/data/browse_comp_debug_set.csv
younik: Done https://github.com/The-AI-Alliance/cube-harness/pull/299/commits/778542f9ba37c6ac28f7cf6d71a7a3e4b2c6f258

cubes/browsercomp/pyproject.toml
younik: Thanks for spotting; cube-web-tool was recently merged, we need to wait it to be on PyPI

younik:
> pytest tests/ exits non-zero (exit code 4) — tests/ directory missing

Added tests in https://github.com/The-AI-Alliance/cube-harness/pull/299/commits/a1cef33a05c14ce1d5a81aaa3f38491be7274d39

> TaskMetadata.extra_info is holding structured per-task fields

Done; they are now lazy loaded

> Bulk data committed inside package source

Removed them; now it is downloaded at install

> No metadata generator script 

Added

> Task.evaluate() makes a stochastic LLM call

Added the full response in info

> No cube-specific unit tests 

Added

> scorer_model default hard-coded in three places

Fixed, now it is set by the caller only, and become a required flag

> authors missing in pyproject.toml

Fixed

> No README.md at the cube package root 

Added

NicolasAG (commented):
# CUBE rc7 migration review — browsercomp-cube

Applying the same PR #118 (`BenchmarkConfig` split) + PR #121 (typed `TaskExecutionInfo`) + PR #123 (`default_tool_config` → `tool_config`) playbook from #322 to this cube. The good news: browsercomp already has a typed `BrowseCompTaskMetadata` and a real `install()` flow, so the migration is mechanical — but it does pull in the `TaskExecutionInfo` work that the web cubes in #322 explicitly opted out of, since `problem`/`answer` are heavy lazy per-task data.

> Reviewed against `cube-standard` `main` (post-rc7: PRs #118, #121, #123 merged).

---

## Blocking — adapts to the rc7 contract

### 1. Split `BrowseCompBenchmark` into `BrowseCompBenchmarkConfig` + runtime `BrowseCompBenchmark`

`src/browsercomp_cube/benchmark.py:24` declares one Pydantic class that holds both ClassVars and runtime hooks. Per rc7 it splits in two — and `__init_subclass__` will reject the config without the new `benchmark_class` ClassVar.

```python
# Before (current)
class BrowseCompBenchmark(Benchmark):
    benchmark_metadata: ClassVar[BenchmarkMetadata] = ...
    task_metadata: ClassVar[dict[str, BrowseCompTaskMetadata]]
    task_config_class: ClassVar[type[TaskConfig]] = BrowseCompTaskConfig
    scorer_model: str
    def _setup(self) -> None: ...
    def close(self) -> None: ...

# After (rc7)
from cube.benchmark import Benchmark, BenchmarkConfig, BenchmarkMetadata

class BrowseCompBenchmark(Benchmark):
    """Plain runtime class — not Pydantic."""
    def _setup(self) -> None: pass    # no shared infra
    def close(self) -> None: pass

class BrowseCompBenchmarkConfig(BenchmarkConfig):
    benchmark_metadata: ClassVar[BenchmarkMetadata] = BenchmarkMetadata(...)
    task_metadata: ClassVar[dict[str, BrowseCompTaskMetadata]]   # auto-loaded
    task_config_class: ClassVar[type[TaskConfig]] = BrowseCompTaskConfig
    benchmark_class: ClassVar[type[Benchmark]] = BrowseCompBenchmark   # required

    scorer_model: str
```

`install` / `uninstall` / `_download_dataset` stay on `BrowseCompBenchmarkConfig` as `@classmethod`s.

### 2. `get_task_configs`: emit `metadata=tm`, not `task_id=tm.id`; rename `default_tool_config` → `tool_config`

`benchmark.py:80-86`. `TaskConfig.task_id` is a derived `@property` in rc7 (`metadata.id`, prefixed with `sub_bench_name` for composites), so callers stamp the `TaskMetadata` instance instead. `default_tool_config` was renamed to `tool_config` on `BenchmarkConfig` (#123).

```python
def get_task_configs(self) -> Generator[BrowseCompTaskConfig, None, None]:
    for tm in self.tasks().values():
        yield BrowseCompTaskConfig(
            metadata=tm,                        # was task_id=tm.id
            tool_config=self.tool_config,       # was self.default_tool_config
            scorer_model=self.scorer_model,
        )
```

`self.tasks()` already returns the (possibly subset) view — no need to walk `task_metadata` directly.

### 3. `BrowseCompTaskConfig.make`: read `self.metadata`, drop the circular-import dodge

`task.py:113-127`. The owning-benchmark lookup goes away because `BenchmarkConfig.get_task_configs()` now stamps `metadata` onto every emitted config.

```python
def make(self, runtime_context=None, container_backend=None) -> BrowseCompTask:
    type(self).verify_installed()    # convention: fail fast if install() hasn't run

    encrypted = self.load_task_execution_info(self.task_id)   # classmethod on TaskConfig
    canary = encrypted[\"canary\"]
    exec_info = BrowseCompExecutionInfo(
        problem=decrypt(encrypted[\"problem\"], canary),
        answer=decrypt(encrypted[\"answer\"], canary),
    )

    tool_cfg = self.tool_config or ToolboxConfig(
        tool_configs=[BraveWebSearchToolConfig(), WebFetchToolConfig(), SubmitAnswerToolConfig()]
    )
    return BrowseCompTask(
        metadata=self.metadata,             # already stamped — no Benchmark lookup
        execution_info=exec_info,
        tool_config=tool_cfg,
        scorer_model=self.scorer_model,
        runtime_context=runtime_context,
        container_backend=container_backend,
    )
```

This deletes the `from browsercomp_cube.benchmark import BrowseCompBenchmark` line and the `BrowseCompBenchmark.task_metadata[self.task_id]` lookup — that was the pre-rc7 circular-import dodge.

### 4. `install()` writes through `cls.task_config_class.task_execution_cache_dir()`

`benchmark.py:50-71`. The cache-dir path moved from `Benchmark` to `TaskConfig` so there's a single source of truth (the path the worker reads via `load_task_execution_info`). The current code calls `cls.task_execution_cache_dir()` directly on the benchmark — that helper isn't on `BenchmarkConfig` in rc7.

```python
@classmethod
def install(cls) -> None:
    exec_cache_dir = cls.task_config_class.task_execution_cache_dir()
    if exec_cache_dir.exists() and any(exec_cache_dir.iterdir()):
        return
    ...

@classmethod
def uninstall(cls) -> None:
    exec_cache_dir = cls.task_config_class.task_execution_cache_dir()
    ...
```

`BenchmarkConfig.install()`'s docstring spells this out: *\"write each task's processed data as a JSON file at `cls.task_config_class.task_execution_cache_dir() / f\"{task_id}.json\"`\"*.

### 5. Promote `problem` / `answer` to a typed `BrowseCompExecutionInfo`

This is what PR #121 is for. The web cubes in #322 didn't need it; browsercomp does, because it ships heavy lazy per-task data through `install()`. Today the decrypted plaintext lives as Pydantic instance fields on `BrowseCompTask` (`task.py:64-65`); rc7's contract puts that on `Task.execution_info: SerializeAsAny[TaskExecutionInfo] | None`.

```python
# task.py
from cube.task import Task, TaskConfig, TaskExecutionInfo, TaskMetadata

class BrowseCompExecutionInfo(TaskExecutionInfo):
    \"\"\"Heavy decrypted payload, populated in TaskConfig.make().\"\"\"
    problem: str
    answer: str

class BrowseCompTask(Task):
    metadata: BrowseCompTaskMetadata          # type: ignore[assignment]
    execution_info: BrowseCompExecutionInfo   # type: ignore[assignment]
    scorer_model: str
    grader_retries: int = 3

    def reset(self):
        self.tool.reset()
        prompt = self.execution_info.problem + _FORMAT_INSTRUCTIONS
        return Observation.from_text(prompt), {\"problem\": self.execution_info.problem}

    def evaluate(self, obs=None):
        submitted = self._submit_tool().last_answer
        if submitted is None:
            return 0.0, {...}
        prompt = _GRADER_TEMPLATE.format(
            question=self.execution_info.problem,
            response=submitted,
            correct_answer=self.execution_info.answer,
        )
        ...
```

The `BrowseCompTask.problem` / `BrowseCompTask.answer` instance fields go away entirely — they're now reached via `self.execution_info`.

(Side note: the README claims *\"ciphertext is only decrypted in `BrowseCompTaskConfig.make()`, so cleartext never lands on disk.\"* That claim isn't really true with either layout — `Task` is JSON-serializable for trajectory recording — but `execution_info` is the rc7-canonical place to put it regardless.)

### 6. Entry point in `pyproject.toml` points at the config class

`cubes/browsercomp/pyproject.toml:24`:

```toml
[project.entry-points.\"cube.benchmarks\"]
browsercomp-cube = \"browsercomp_cube.benchmark:BrowseCompBenchmarkConfig\"
```

The CLI (`cube list` / `cube test`) loads the config class and calls `.make(infra)` — the runtime `Benchmark` is never the entry point in rc7.

### 7. Pin `cube-standard>=0.1.0rc7`

`cubes/browsercomp/pyproject.toml:18`. The PR body says it depends on `cube-standard` PR #107, but the contract this needs is rc7 (which lands #118 + #121 + #123). Without the pin, `pip install` may resolve an older release that doesn't have `BenchmarkConfig` / `TaskExecutionInfo` and the cube simply won't import.

```toml
dependencies = [
    \"cube-standard>=0.1.0rc7\",
    \"cube-web-tool\",
    \"litellm\",
]
```

---

## Debug + tests follow-ups

### `debug.py`

Mirrors the same split:

```python
class DebugBrowseCompTask(BrowseCompTask):
    def _call_grader(self, prompt, scorer_model):
        submitted = self._submit_tool().last_answer or \"\"
        ok = self.execution_info.answer.lower() in submitted.lower()
        return ok, f\"correct: {'yes' if ok else 'no'} (debug grader)\"

class DebugBrowseCompTaskConfig(BrowseCompTaskConfig):
    def make(self, runtime_context=None, container_backend=None):
        idx = int(self.metadata.id.rsplit(\"-\", 1)[-1])
        record = _DEBUG_RECORDS[idx]
        return DebugBrowseCompTask(
            metadata=self.metadata,
            execution_info=BrowseCompExecutionInfo(problem=record[\"problem\"], answer=record[\"answer\"]),
            tool_config=self.tool_config or ToolboxConfig(tool_configs=[SubmitAnswerToolConfig()]),
            scorer_model=self.scorer_model,
            container_backend=container_backend,
        )

class DebugBenchmark(BrowseCompBenchmark):
    \"\"\"Runtime — no-op setup/close.\"\"\"

class DebugBrowseCompBenchmarkConfig(BrowseCompBenchmarkConfig):
    task_metadata: ClassVar[dict[str, BrowseCompTaskMetadata]] = _debug_task_metadata()
    task_config_class: ClassVar[type[TaskConfig]] = DebugBrowseCompTaskConfig
    benchmark_class: ClassVar[type[Benchmark]] = DebugBenchmark
    scorer_model: str = \"debug-grader-unused\"

    @classmethod
    def install(cls) -> None: pass     # debug benchmark uses inline records
    @classmethod
    def uninstall(cls) -> None: pass

def get_debug_benchmark(infra=None) -> Benchmark:
    return DebugBrowseCompBenchmarkConfig().make(infra)
```

The `infra=None` parameter on `get_debug_benchmark` matches the `cube.testing` contract used elsewhere (#322 calls this out for every web cube).

### Tests

Drop in a config round-trip — that's the new acceptance bar from PR #118 because it's what catches `SerializeAsAny` regressions:

```python
def test_config_round_trip():
    cfg = BrowseCompBenchmarkConfig(scorer_model=\"gpt-5.4-mini\")
    rehydrated = BrowseCompBenchmarkConfig.model_validate_json(cfg.model_dump_json())
    assert rehydrated == cfg
```

A second one for `BrowseCompTaskConfig` (with `metadata=...` set) catches `TaskMetadata` subclass-stripping if `SerializeAsAny` is missing somewhere.

---

## What does *not* change

- `task_metadata.json` location — still loaded by `__init_subclass__` from next to `benchmark.py`. ✓
- `BrowseCompTaskMetadata` itself — already a typed subclass, already public-only. ✓
- `tags=[\"web\", \"browser\", \"reasoning\", \"nlp\"]` — already registry-conformant. ✓
- `_DATASET_URL`, `crypto.py`, `tool.py`, `SubmitAnswerTool` — untouched. ✓

---

## Suggested commit order

1. Pin `cube-standard>=0.1.0rc7` in `pyproject.toml`.
2. Split `BrowseCompBenchmark` (config + runtime), wire `benchmark_class`.
3. Move `install` / `uninstall` to use `cls.task_config_class.task_execution_cache_dir()`.
4. Introduce `BrowseCompExecutionInfo`; make `BrowseCompTask` read `self.execution_info.{problem,answer}`.
5. Migrate `BrowseCompTaskConfig.make` to use `self.metadata` + `self.load_task_execution_info`.
6. Migrate `debug.py` (split + `infra=None` signature).
7. Add config round-trip tests.
8. Bump entry point to `BrowseCompBenchmarkConfig`.

Each step is independently testable with `cube test browsercomp-cube` against the debug suite.

younik:
@NicolasAG Done it in https://github.com/The-AI-Alliance/cube-harness/pull/299/commits/fdeaec5ede1231314e519ab40248b278d46cef0c

NicolasAG:
Picked up the review notes — landing in two commits on this branch:

**1. Recipe + README fixes + adopt `Task[TMeta]` / `Benchmark[Config]` generics.**
- `recipes/browsercomp.py` was constructing `BrowseCompBenchmark(default_tool_config=..., scorer_model=...)` and passing the resulting instance as `benchmark=...` — neither matches the actual signatures, so the recipe `TypeError`d immediately. Switched to `BrowseCompBenchmarkConfig(...)` + `benchmark_config=` (matches `recipes/hello_miniwob.py`).
- Two more recipe-path bugs surfaced once the recipe ran: `BrowseCompTaskConfig.make()` was calling `type(self).verify_installed()` (instance method, missing `self`) and `self.load_task_execution_info(self.task_id)` (the helper takes no positional args — uses `self.task_id` internally).
- Parameterised `BrowseCompTask(Task[BrowseCompTaskMetadata])` and `BrowseCompBenchmark(Benchmark["BrowseCompBenchmarkConfig"])`; replaced the `execution_info: ... # type: ignore[assignment]` field redeclaration with a typed `_exec` property (same pattern as `swebench-verified-cube`). Drops both `# type: ignore` lines.
- Fixed three malformed inline links in `README.md`.
- Bumped `cube-standard>=0.1.0rc8`.

**2. `named_subsets` keyed by `topic`.**
The dataset has 10 well-balanced topics (smallest `Politics` @ 59 tasks, largest `TV shows & movies` @ 205, no empty values). Per-topic accuracy is the standard slicing for retrieval/QA benchmarks, so surfacing them as named subsets is cheap and useful. Coming in commit 2.

NicolasAG (approved):

PR #301, opened 2026-04-20
fix(webarena-verified): defer setup validation from __init__ to _setup()

NicolasAG:
## Summary

- Moved the `_validate_setup_mode` model validator out of `__init__` (where it blocked no-arg instantiation) and into `_setup()` where it belongs
- Validation and infra/manual mode dispatch now run only when the benchmark is actually being used, not at construction time
- Removed unused `model_validator` import; updated docstring

## Why

The registry `quick_check.py` introspects the `Benchmark` class (reads `benchmark_metadata.num_tasks`, `task_config_class` etc.) without setting up any Docker infra. The old `@model_validator` raised at construction time, preventing even static metadata reads.

Fixes: https://github.com/The-AI-Alliance/cube-registry/pull/33

## Test plan
- [ ] `uv run python -c "from webarena_verified_cube.benchmark import WebArenaVerifiedBenchmark; b = WebArenaVerifiedBenchmark(); print(b.benchmark_metadata.num_tasks)"` prints 812
- [ ] Passing `infra=` without `resources=` still raises at `setup()` time
- [ ] Passing both `wav_config.environments` and `infra=` still raises at `setup()` time

### DIFF
--- a/cubes/webarena-verified/src/webarena_verified_cube/benchmark.py
+++ b/cubes/webarena-verified/src/webarena_verified_cube/benchmark.py
@@ -4,7 +4,7 @@ import urllib.request
 from collections.abc import Generator
 from typing import ClassVar
 
-from pydantic import PrivateAttr, model_validator
+from pydantic import PrivateAttr
 from cube.benchmark import Benchmark, BenchmarkMetadata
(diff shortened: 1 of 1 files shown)


younik (approved):
LGTM!

PR #309, opened 2026-04-23
feat: meta-agent scaffolding — task hints and episode task_id wiring

recursix:
## Summary

Minimal scaffolding for the meta-agent iteration loop:

- `GennyConfig.task_hints: dict[str, str]` — maps task_id → hint text injected into the agent's context after the goal. Also adds a general `hint: str` field for subset-wide hints.
- `Genny.__init__` accepts `task_id: str | None`; resolves hint from config at init time.
- `Episode.run()` passes `task_id=self.config.task_id` to `agent_config.make()`, threading the task identity from benchmark config down to the agent.
- `.claude/commands/meta-agent.md`: the meta-agent skill documenting the orient→select→run→analyse→fix→validate→log loop.

This is the smallest useful unit — it enables per-task hint injection without any benchmark-specific code. WorkArena findings and structural improvements follow in separate PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #153, opened 2026-02-19
Xray Agent Comparison task-wise


recursix:
allow comparing 2 loaded agents side by side, to see which tasks succedd vs which task fail and allow filtering based on A > B or B < A or B & A fails or B & A succeeds …

PR #310, opened 2026-04-23
feat(workarena): WorkArena L1 findings — tools, hints, browsergym actions

recursix:
## Summary

WorkArena L1 improvements from meta-agent iteration sessions. Stacks on #309.

- **cubes/workarena/task.py**: remove \`filter_actions\`, always include \`WorkArenaInfeasibleTool\`
- **agents/genny.py**: add \`task_precision\` dict — per-task goal-level details injected before the hint
- **tools/browsergym.py**:
  - \`tool_action\` dispatch layer — custom \`@tool_action\` methods exposed as agent actions
  - \`keyboard_type_into\` — fires keyboard events character-by-character for autocomplete fields
  - \`submit_form\` — calls \`gsftSubmit()\` directly in \`gsft_main\` iframe
  - \`js_eval\` — evaluates JS in the browser
  - \`axtree_with_visible\` / \`axtree_with_clickable\` / \`action_error_full_traceback\` config flags
  - Performance: skip detached frames in \`_wait_dom_loaded\`, skip \`extract_dom_extra_properties\` when flags are off
- **meta_agent/**: \`workarena_hints.py\`, \`recipes/workarena_l1_full.py\`, journal entries (4 sessions)

### CI / infra fixes
- **cube-ci-fast.yml**: add \`--no-reset-check\` to both \`cube test\` calls (flag lands in cube-standard #120)
- **Makefile**: switch \`lint\`/\`lint-check\` from \`uv run ruff\` to \`uvx ruff\` — ruff runs in an isolated tool env, no project venv needed, so CI doesn't fail when editable local paths are absent
- **pyproject.toml / uv.lock**: remove \`cube-browser-playwright\` from \`[tool.uv.sources]\` — it's published via cube-standard's release cycle; locking to a local editable path broke CI after the last \`uv lock\` regeneration

## Test plan
- [ ] \`keyboard_type_into\` works on ServiceNow autocomplete fields
- [ ] \`submit_form\` writes localStorage and submits create-record forms correctly
- [ ] \`WorkArenaInfeasibleTool\` is always available in the action set
- [ ] \`make lint-check\` passes in CI (ruff workflow green)


amanjaiswal73892:
Three issues that need to be addressed before merging:

**1. `submit_form` is missing from the diff (blocker)**

`_CREATE_PRECISION` tells the agent to call `submit_form()` to submit forms, and `_CREATE_HINT` references it as the correct submission method. But there's no `submit_form` implementation in `web_actions.py` or anywhere else in the diff. The journal describes its development in detail across multiple iterations, so it clearly exists somewhere — but it didn't make it into this PR. Agents will be instructed to call an action that doesn't exist.

**2. `task_precision` vs `task_clarification` field name mismatch**

`workarena_hints.py` exports `WORKARENA_TASK_PRECISION` and its module docstring shows:
```python
agent_config = GennyConfig(
    ...
    task_precision=WORKARENA_TASK_PRECISION,
)
```
But `GennyConfig` (from #309) has `task_clarification`, not `task_precision`. Either the field was renamed during #309 and `workarena_hints.py` wasn't updated, or a new `task_precision` field needs to be added to `GennyConfig`. Either way, `WORKARENA_TASK_PRECISION` is currently unused — it's never passed to `GennyConfig` in the recipe.

**3. `uv.lock` pins `cube-standard` to a local editable path**

`pyproject.toml` adds `cube-standard = { path = "../cube-standard", editable = true }` to `[tool.uv.sources]`, and the lockfile reflects this. This assumes `cube-standard` is checked out as a sibling directory, which will break CI and any developer whose layout differs. The PR description notes that `cube-browser-playwright` was *removed* from local sources for exactly this reason — this looks like the same problem applied to `cube-standard`.

recursix:
Thanks for the thorough review Aman. Addressed below.

**1. submit_form**

`submit_form` was built early in development (April 10) as the way to trigger the correct ServiceNow form action. It was later superseded by three WorkArena-side patches in `form.py` (PR #153 on WorkArena) that intercept `sysverb_insert_and_stay` at the JS and network layers — so the agent can now click the visible Submit button directly and the correct verb always reaches the server. The journal entries document the development history including that path, which is why it reads as if `submit_form` should be present.

Removed the stale `submit_form` mention from the recipe docstring. 100% on all 5 create tasks across 5 seeds confirms the current approach works without it.

**2. task_precision vs task_clarification**

Fixed — the module docstring now shows `task_clarification`. The recipe was already wiring it correctly, only the docstring was stale.

**3. cube-standard local editable path**

Removed from `[tool.uv.sources]` and regenerated the lockfile. `cube-standard` now resolves from PyPI (`0.1.0rc6`).

Also removed `test_workarena_cheat.py` — it was a diagnostic script for debugging the sysverb issue, not a real test.

meta_agent/journal/2026-04-08_workarena-l1-baseline.md
younik, line 1: have you commit those on purpose? Do you want to merge them?
They are report of the meta-agent that can be useful while reviewing the fixes, but I wouldn't merge them in the repo

src/cube_harness/tools/browsergym.py
```diff
@@ -299,6 +323,16 @@ def _bgym_obs_to_cube_obs(self, bgym_obs: dict[str, Any]) -> Observation:
...
 
 
+# Descriptions that replace BrowserGym's upstream text.
+# Use sparingly — only when the upstream description is misleading about when to use the action.
+_ACTION_DESCRIPTION_OVERRIDES: dict[str, str] = {
+    "fill": (
+        "Fill a form field by setting its value directly. It does not fire keyboard events, so autocomplete suggestions"
+        " will not appear. Use keyboard_type_into() instead for fields that require autocomplete."
+    ),
+}
```
younik, line 333: I feel this shouldn't be here; this is the browsergym tool, so we are expecting browsergym action. For example, if a future version of BGym will support keyboard event firing, this override will be wrong. 

If these are necessary to get good perf (which shouldn't, if we want to match BrowserGym perf), then this should be another tool

meta_agent/workarena_hints.py
younik, line 1: should this go inside the workarena cube instead?

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
 # XRay trajectory viewer
 analyze = [
     "gradio~=5.49",
+    "pandas~=2.0",
```
younik, line 39: I don't see pandas used anywhere; shouldn't be a dep of whole cube

src/cube_harness/tools/web_actions.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self, browser: BrowsergymTool) -> None:
+        self._browser = browser
+
+    def execute_action(self, action: Action) -> Observation | StepError:
+        method = self.get_action_method(action)
+        try:
+            result = str(method(**action.arguments) or "Success")
+        except Exception as e:
+            return StepError.from_exception(e)
```
younik, line 41: Should be narrower than Exception, so real bugs surface

younik: A Playwright Error is a better replacement

src/cube_harness/tools/browsergym.py
```diff
@@ -157,12 +163,15 @@ def _wait_dom_loaded(self) -> None:
...
+                page.wait_for_load_state("domcontentloaded", timeout=1500)
             except Error:
                 pass
             for frame in page.frames:
+                # un necessary to wait for detached frames, and waiting on them raises a timeout error, so skip them
+                if frame.is_detached():
+                    continue
                 try:
-                    frame.wait_for_load_state("domcontentloaded", timeout=3000)
+                    frame.wait_for_load_state("domcontentloaded", timeout=1500)
```
younik, line 174: is there a reason for these changes? If yes, maybe document for future reference; we may also want to make these settable

tests/test_browsergym.py
```diff
@@ -589,3 +593,80 @@ def test_close_noop_when_no_browser(self) -> None:
...
+        li.textContent = e.key;
+        document.getElementById('suggestions').appendChild(li);
+      }}
+    }});
+  </script>
+</body></html>""".format(bid_attr=BROWSERGYM_ID_ATTRIBUTE, bid=_AUTOCOMPLETE_BID)
+
+
+@pytest.fixture(scope="module")
+def live_page():
```
younik, line 623: missing return type

tests/test_browsergym.py
```diff
@@ -589,3 +593,80 @@ def test_close_noop_when_no_browser(self) -> None:
...
+        live_page.set_content(_AUTOCOMPLETE_PAGE)
+
+        get_elem_by_bid(live_page, _AUTOCOMPLETE_BID).press_sequentially("hello", delay=0)
+
+        keydown_count = live_page.evaluate("window.keydownCount")
+        assert keydown_count == 5, "press_sequentially must fire one keydown per character"
+        suggestions = live_page.locator("#suggestions li").count()
+        assert suggestions == 5
+
+    def test_keyboard_type_into_action_fires_keydown_events(self, live_page) -> None:
```
younik, line 657: missing type for live_page

tests/test_browsergym.py
```diff
@@ -589,3 +593,80 @@ def test_close_noop_when_no_browser(self) -> None:
...
+        const li = document.createElement('li');
+        li.textContent = e.key;
+        document.getElementById('suggestions').appendChild(li);
+      }}
+    }});
+  </script>
+</body></html>""".format(bid_attr=BROWSERGYM_ID_ATTRIBUTE, bid=_AUTOCOMPLETE_BID)
+
+
+@pytest.fixture(scope="module")
```
younik, line 622: do we need scope here? making a page object shouldn't be so expensive; we risk to have failures cause by other tests that are then misleading

src/cube_harness/tools/web_actions.py
```diff
@@ -0,0 +1,102 @@
...
+            else:
+                target = next((f for f in self._browser.page.frames if f.name == frame), None)
+                if target is None:
+                    return f"Failed: frame {frame!r} not found"
+            raw = target.evaluate(
+                f"() => {{ try {{ return {code}; }} catch(e) {{ return 'JS error: ' + e.message; }} }}"
+            )
+            return json.dumps(raw, default=str)
+        except Exception as e:
+            return f"Failed: {type(e).__name__}: {e}"
```
younik, line 87: too broad, it will hide real bugs; use Playwright Error, at least

src/cube_harness/tools/web_actions.py
```diff
@@ -0,0 +1,102 @@
...
+
+class ExtraWebActionsTool(Tool):
+    """Extra browser actions that complement BrowserGym's built-in set.
+
+    Holds a reference to a BrowsergymTool to share its page and observation
+    extraction — both tools operate on the same browser session.
+    """
+
+    def __init__(self, browser: BrowsergymTool) -> None:
+        self._browser = browser
```
younik, line 34: Is ExtraWebActionsTool intended as an extension of BrowsergymTool only? If so, it should likely extend it
If not, then the browser at init can be a generic BrowserTaskTool (but keep in mind it should go in cube-browser-tool asap)

meta_agent/recipes/workarena_l1_full.py
```diff
@@ -0,0 +1,167 @@
...
+            ExtendedBrowserConfig(
+                browser=BrowsergymConfig(
+                    browser=PlaywrightSessionConfig(headless=headless, timeout=60000),
+                    use_screenshot=True,
+                    use_axtree=True,
+                    use_html=False,
+                    axtree_with_clickable=True,
+                    axtree_with_visible=True,
+                ),
+            ),
```
younik, line 78: The only relevant difference I see compared to workarena recipes is the use of this tool that has extended actions (and the flags of BGym); maybe we can integrate this in the workarena recipe and use args to decide what to use.

meta_agent/workarena_hints.py
```diff
@@ -0,0 +1,226 @@
...
+    "workarena.servicenow.filter-asset-list": _FILTER_HINT,
+    "workarena.servicenow.filter-change-request-list": _FILTER_HINT,
+    "workarena.servicenow.filter-hardware-list": _FILTER_HINT,
+    "workarena.servicenow.filter-incident-list": _FILTER_HINT,
+    "workarena.servicenow.filter-service-catalog-item-list": _FILTER_HINT,
+    "workarena.servicenow.filter-user-list": _FILTER_HINT,
+}
+
+# Backward compat
+WORKARENA_DEFAULT_HINT: str = ""
```
younik, line 226: This is unused; if so, remove

younik (approved):
Approving to iterate fast, but have a look at the comment, thanks!

meta_agent/journal/2026-04-08_workarena-l1-baseline.md
recursix: I'm planning to accumulate reports. Agree that it's not 100% it's place in there, but it's a sub dir journal. I think it can be useful long term

src/cube_harness/tools/browsergym.py
recursix: 3s was always too much. If the page is not ready after this, the agent should notice and call noop.

src/cube_harness/tools/browsergym.py
recursix: That's a good point

src/cube_harness/tools/web_actions.py
recursix: At the moment it's an extention to bgym tool. Default would be not to use it. For understanding what actions is missing, it would be useful for development phase

PR #311, opened 2026-04-23
feat: structural improvements — agent_name, episode cleanup, trace framework

recursix:
## Summary

Infrastructure and observability improvements. Stacks on #310.

- **agent_name property**: added to `AgentConfig`, `GennyConfig`, `ReactAgentConfig`, `LegacyGenericAgentConfig` — used in trajectory metadata instead of `type(config).__name__`
- **episode.py**: use `agent_name`, include `action_schemas` in trajectory metadata, remove validator feedback from agent observations, logging improvements
- **analyze/trace.py**: new trace analysis framework
- **scripted_episode.py**: scripted episode support for replay/debugging
- **analyze/inspect_results.py**: expanded analysis helpers
- **exp_runner.py**: demote Ray dashboard warning to DEBUG
- **experiment.py**: retry logic improvements
- **metrics/tracer.py**: pass LLM credentials to Ray workers, load `.env` in worker processes
- **pyproject.toml**: dependency updates

## Test plan
- [ ] `agent_name` appears correctly in saved trajectory metadata
- [ ] `action_schemas` are recorded in trajectory metadata
- [ ] Validator feedback no longer leaks into agent observations
- [ ] Ray workers receive LLM credentials correctly


amanjaiswal73892:
Four things that need a decision before merging:

**1. `task_precision` vs `task_clarification` (carried from #310)**

`workarena_hints.py` exports `WORKARENA_TASK_PRECISION` and its docstring shows `task_precision=WORKARENA_TASK_PRECISION` passed to `GennyConfig`. But `GennyConfig` has `task_clarification`, not `task_precision`. In the recipe, the line is commented out: `# task_precision=WORKARENA_TASK_PRECISION,`. So `WORKARENA_TASK_PRECISION` is currently dead — it's imported but never applied. Pick one name and align everywhere.

**2. `keyboard_type_into` / `js_eval` duplicated across two tools**

Both `BrowsergymTool` (`browsergym.py`) and `ExtraWebActionsTool` (`web_actions.py` from #310) implement `keyboard_type_into` and `js_eval`, with slightly different internals. The recipe in this PR uses plain `BrowsergymConfig` (not `ExtendedBrowserConfig`), which suggests the intent is to consolidate into `BrowsergymTool` — but `web_actions.py` still exists. One of them should win; the other should be removed or explicitly deprecated.

**3. `env_output` in `scripted_episode.py` — side-effect or dead code?**

```python
env_output: EnvironmentOutput = EnvironmentOutput(  # noqa: F841
    obs=tool.page_obs() if hasattr(tool, "page_obs") else Observation(),
    ...
)
```

The variable is never read. The only effect is calling `tool.page_obs()` after `task.setup()`. Is this warm-up intentional, or is it leftover scaffolding? If intentional, it should just be `tool.page_obs()` with no assignment. If not needed, remove it entirely.

**4. `_get_frame_for_bid` re-implements BrowserGym internals**

The new `_get_frame_for_bid` method in `BrowsergymTool` tries to manually resolve BID prefixes to frames, while `web_actions.py` uses the official `get_elem_by_bid(page, bid)` helper from BrowserGym. Worth checking if `get_elem_by_bid` can be used in `keyboard_type_into` here too, to avoid the custom re-implementation diverging from BrowserGym's actual encoding.

amanjaiswal73892:
Found and fixed three bugs introduced in this PR (commit 2e4c4276):

1. **`GennyConfig.make()` — reverted safety fix** (`agents/genny.py`): Restored the explicit `task_id: str | None = None` parameter. The `kwargs.get("task_id")` approach silently swallows misspelled keyword args, which commit c060ab6a had specifically fixed.

2. **`format_agent_comparison` — wrong values on reordered DataFrames** (`analyze/inspect_results.py`): The `{name: str(df.iloc[i][var]) for i, name in enumerate(agent_names)}` pattern assumed row `i` corresponds to `agent_names[i]`. Replaced with `df.iterrows()` so values are always tied to their actual row. Added a regression test.

3. **`extra_metadata` merge order** (`episode.py`): `**env_output.info` was unpacked after `**(extra_metadata or {})`, so environment keys silently overwrote caller-supplied metadata. Swapped the order so `extra_metadata` wins on collision.

PR #314, opened 2026-04-27
feat: introduce ~/.env-cube credential convention

recursix:
Just a test to bring a broader discussion of how we handle env vars in general

## Summary

- Add `~/.env-cube` as the single, scope-isolated credential file for cube recipes
- Ray workers call `load_dotenv(Path.home() / ".env-cube", override=False)` on startup, so credentials flow naturally on same-machine runs (env is inherited) and work on multi-node clusters where each worker machine needs its own credential file
- `get_trace_env_vars()` now has a docstring explicitly documenting that it forwards only OTel tracing vars, not secrets
- README updated to show the `~/.env-cube` pattern instead of `export OPENAI_API_KEY`

## Why a separate file

Using `~/.env` risks loading unintended variables from a global shell env file. Using the repo-local `.env` ties credentials to repo layout. `~/.env-cube` is explicit, scope-isolated, and works symmetrically for local and multi-node setups.

## Test plan

- [ ] Verify `make debug` works with credentials in `~/.env-cube` (no shell export needed)
- [ ] Verify Ray worker still gets credentials on a local run (env inheritance, `load_dotenv` is no-op)
- [ ] Smoke-test multi-node: place `~/.env-cube` on worker, confirm it picks up credentials


amanjaiswal73892 (commented):
The worker-side `load_dotenv` works correctly (it runs inside the Ray actor and sets env vars in that worker process), but we already have a better mechanism for this: `runtime_env["env_vars"]` at `ray.init()` pushes vars from the driver to every worker automatically. Adding a symmetric `load_credential_vars()` helper alongside `get_trace_env_vars()` and merging both into `runtime_env` would eliminate the per-node file requirement entirely — credentials loaded once on the driver, Ray handles distribution. No `~/.env-cube` needed on worker machines.

___
The above is optional: the PR is functional as is and solves a important issue for multi-node ray jobs.

src/cube_harness/exp_runner.py
```diff
@@ -58,6 +60,9 @@ def _run_with_ray_impl(
 
     @ray.remote
     def run_episode(episode: Episode) -> Trajectory:
+        # Same-machine workers inherit the parent process env (no-op here).
+        # Multi-node workers need credentials on their own machine — place them in ~/.env-cube.
+        load_dotenv(Path.home() / ".env-cube", override=False)
```
conte91, line 65: This will likely get replicated across multiple modules. I wonder if it makes sense to create a shared  `load_cube_env` helper (just to make sure all callers are in sync, there's clearly not much value in the "helper" bit itself since it's literally 1 LoC)

README.md
```diff
@@ -29,12 +29,14 @@ make install
 
 ### API Keys
 
-Set your OpenAI API key:
+Create `~/.env-cube` with your credentials:
```
conte91, line 32: As mentioned yesterday: Running `cube` already has the unexpected result of polluting the user's home directory with a `cube_harness_results` dir. Creating more can be annoying, on a computer which is more likely than not used by a desktop user doing mostly work _other_ than CUBEs; I wonder if we should just follow the [XDG basedir](https://specifications.freedesktop.org/basedir/latest/) desktop standard here - just like [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard), it allows people to know where all "things" for a "program" live, and it's well supported by the community (e.g. https://pypi.org/project/xdg-base-dirs/). I believe this is a good moment to start thinking about unifying where our configs and artifacts live (benchmark install artifacts is also another one that falls in the "pollute user state" category). The bonus point is, if we follow this it's easier for somebody to "reset" their state because they'll know they can delete ~/.config/cube or the likes and not have any surprising side effect.

meta_agent/recipes/workarena_l1_full.py
```diff
@@ -27,11 +27,8 @@
 # meta_agent/ is not a Python package — add it to sys.path so we can import workarena_hints.
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
 
-# Load .env so credentials are available even when the shell didn't source ~/.zshrc.
-# Ray workers inherit the parent process env, so this must run before ray.init().
-# Falls back to ~/.env for developers who keep credentials in a global env file.
-_project_env = Path(__file__).resolve().parents[2] / ".env"
-load_dotenv(_project_env if _project_env.exists() else Path.home() / ".env", override=True)
+# Load credentials from ~/.env-cube before ray.init() so Ray workers inherit them.
+load_dotenv(Path.home() / ".env-cube")
```
conte91, line 31: If this is part of the cube standard: we should probably not have to type this in every recipe. maybe it should be the benchmark runner which loads the env on startup, WDYT?

conte91, line 31: Bonus point on the "helper": on at least 2 enterprise scenarios I've personally met, leaving API keys on employee's home directories in env files is strictly forbidden. We might need to support different modes of loading the env depending on a config setting (that one can live in the env :P)

Pydantic is really good as a general "configuration manager" - it can read and aggregate from a lot of sources (including custom ones, secrets, keyrings, you name it). this is why I was pushing back on using env variables as an accepted pattern and trying to relegate it to 'last resort' situations where that choice is out of our control (e.g. webarena requiring env variables to be set - that can be done by the WA cube wrapper at the boundary). Trying to pydantic-normalize everything at the input and output boundaries will make it trivial to change the way that these IO values must be provided should we need to.

This is more food for thought than an actual immediate suggestion, just something i wanted to explain better since yesterday's discussion was rushed and i didn't give enough context

conte91 (commented):
Thanks! Please consider my comments more of a conversation/design suggestion than an immediate blocker, this PR is already a step in the right direction no matter what the bigger picture is so i have no problems with merging it :)

Please consider whether we can have a single ingestion point for environment to make sure that "recipe" writers don't have to remember this (otherwise the whole "standard" risks becoming more of a "suggestion" and people will still need to write everything themselves!)

PR #315, opened 2026-04-27
rfc: episode status file for robust resume/retry

recursix:
## Problem

The current retry logic scans full trajectories to decide what to re-run. Three failure modes:

1. **Silent crash** — if a Ray worker dies (OOM, force-cancel, machine failure), no terminal status is written. The episode is silently dropped, neither retried nor counted as done.
2. **Recovered errors cause false retry** — a StepError the harness absorbed mid-episode still flags the trajectory for retry even though the loop completed normally.
3. **Status requires reading the payload** — deserializing thousands of trajectories to check retry eligibility is expensive and fragile.

## Proposed solution

Write a small `status.json` in each episode directory. Status is ground truth for retry; trajectory files are data.

**Statuses:** `RUNNING` → `COMPLETED` / `FAILED` / `CANCELLED`

**Heartbeat:** a background thread updates `last_heartbeat_at` every 30s. A stale heartbeat on a `RUNNING` episode = dead worker = eligible for retry. No Ray dashboard dependency.

**retry_count:** tracked in `status.json`, capped at `max_retries` (default 3) to prevent infinite loops.

The retry check becomes: read one small JSON file per episode directory. No trajectory deserialization.

## Files

- `openspec/changes/episode-status/proposal.md` — rationale, schema, alternatives considered
- `openspec/changes/episode-status/deltas.md` — precise ADDED/MODIFIED/REMOVED against episode, experiment, and storage specs

## Open questions for review

1. `max_retries` default: 3?
2. Heartbeat interval 30s / staleness threshold 2min?
3. Should `CANCELLED` be retried by default or only on explicit opt-in?

### DIFF
--- a/openspec/changes/episode-status/deltas.md
+++ b/openspec/changes/episode-status/deltas.md
@@ -61,7 +61,7 @@ directory. No trajectory deserialization in the retry decision path.
 Controls how many times a failed episode is retried. Episodes at `retry_count >= max_retries`
 are reported as permanently failed and excluded from future runs.
 
-### ADDED — `max_retry_rounds: int = 1` parameter on `run_with_ray` and `run_sequentially`
+### ADDED — `max_retry_rounds: int = 3` parameter on `run_with_ray` and `run_sequentially`
 
 After the main run completes, the runner checks for retriable episodes. If any exist
 and `retry_rounds < max_retry_rounds`, it runs again with `retry_failed=True` on the
--- a/openspec/changes/episode-status/deltas.md
+++ b/openspec/changes/episode-status/deltas.md
@@ -128,11 +128,6 @@ Run at:
 - The start of `get_episodes_to_run()` when `resume=True`
 - After `ray.shutdown()` in `_run_with_ray_impl`
 
-### ADDED — `is_retriable(status, max_retries) -> bool` (in `experiment.py`)
-
-Standalone helper; `True` if status is `None` or in `RETRIABLE_STATUSES` with
-`retry_count < max_retries`.
-
 ### ADDED — `max_retries: int = 3` field on `Experiment`
 
 Controls how many times a failed episode is retried. Episodes at `retry_count >= max_retries`
@@ -160,6 +155,9 @@ def write_episode_status(self, trajectory_id: str, status: EpisodeStatus) -> Non
 
 def read_episode_status(self, trajectory_id: str) -> EpisodeStatus | None: ...
     # Returns None if status.json does not exist or is corrupt
+
+def archive_episode(self, trajectory_id: str) -> None: ...
+    # Renames the episode directory to <id>.archived_<ts>/ preserving per-attempt history
 ```
 
 ### ADDED — `FileStorage` implementation
@@ -183,8 +181,9 @@ Status files live at `episodes/{trajectory_id}/status.json`.
 Before submitting episodes to Ray, the driver calls `_pre_claim(storage, episode)`:
 
 - Reads any prior `status.json`.
-- If prior is terminal, archives the episode directory (preserving per-attempt history).
+- If prior is terminal, calls `storage.archive_episode(traj_id)` (public `Storage` Protocol method) to preserve per-attempt history.
 - Writes `status=QUEUED` with `last_heartbeat_at=None` and `retry_count` from `next_retry_count(prior)`.
+- Only called for episodes that will actually run — in `debug_limit` mode, episodes beyond the limit are not pre-claimed.
 
 A concurrent runner sees `QUEUED` / `RUNNING` and skips those episodes.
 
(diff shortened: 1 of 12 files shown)


manuel-delverme:
This makes sense!

Shailesh-05 (commented):
Looks good overall— I validated the core implementation and tests (`uv run pytest -q tests/test_experiment.py tests/test_storage.py` passed), and the runtime behavior now looks good (status-driven retry, no Ray dashboard dependency, retry cap support, stale sweep, etc.).

there were 2 doc/spec mismatches that had to be fixed which I added:

1. **`openspec/changes/episode-status/deltas.md` still described a background heartbeat thread**, but implementation used **step-boundary heartbeats in the worker main thread** (`Episode._run_loop` writes heartbeat before each turn).  
  So,I updated the delta text to match actual behavior (and RFC proposal language).

2. **`deltas.md` still describes resume/retry semantics with `retry_failed` flags**, but current API is now driven by `resume` + status-based selection + auto-retry in runners (`max_retry_rounds`).  
  So, aligned the semantics table/wording to current public API so future readers don’t get misled.

Seeing if anything else is missing or needs to be modified.

amanjaiswal73892:
### Code review [Issues addressed in follow-up]

Found 4 issues :

1. `_run_sequentially_impl` pre-claims **all** episodes as `QUEUED` before applying `debug_limit`, leaving un-run episodes orphaned. `QUEUED` is not in `RETRIABLE_STATUSES`, so `_should_relaunch` returns `False` for them and they are silently skipped on the next `resume=True` run until `orphan_threshold_s` (1 hour default) elapses. Fix: move the `_pre_claim` loop to after the slice.

https://github.com/The-AI-Alliance/cube-harness/blob/0cb169139a586eb7345b85f74ab76d02c2692256/src/cube_harness/exp_runner.py#L436-L442

2. `_poll_ray` no longer calls `storage.save_failure(traj_id, tb)` when `ray.get()` raises, so `failure.txt` is never written. `xray_utils.trajectory_status()` classifies a trajectory as `"system_error"` only when `_failure_text` is present (populated from `failure.txt`); without it, Ray-level crashes show as `"queued"` or `"running"` in XRay and the stack trace tab is empty.

https://github.com/The-AI-Alliance/cube-harness/blob/0cb169139a586eb7345b85f74ab76d02c2692256/src/cube_harness/exp_runner.py#L263-L267

https://github.com/The-AI-Alliance/cube-harness/blob/0cb169139a586eb7345b85f74ab76d02c2692256/src/cube_harness/analyze/xray_utils.py#L58-L63

3. `_pre_claim` calls `storage._episode_dir()` and `storage._archive_episode()` — underscore-private methods of `FileStorage` that are not on the `Storage` Protocol. Callers outside the class that type against `Storage` would break, and this couples the runner to `FileStorage`'s internal layout. A public `archive_episode(trajectory_id)` method on `FileStorage` (and optionally `Storage`) would fix this.

https://github.com/The-AI-Alliance/cube-harness/blob/0cb169139a586eb7345b85f74ab76d02c2692256/src/cube_harness/exp_runner.py#L43-L48

4. `is_retriable()` in `experiment.py` is defined but never called — `_should_relaunch()` contains identical logic. Dead code that duplicates the retry-eligibility check (CC-005 minimalist imperative).

https://github.com/The-AI-Alliance/cube-harness/blob/0cb169139a586eb7345b85f74ab76d02c2692256/src/cube_harness/experiment.py#L241-L247

amanjaiswal73892:
### Follow-up: issues resolved

All four issues from the earlier review are addressed in 9869c29.

1. **`debug_limit` QUEUED orphans** — `_pre_claim` loop moved to after the `debug_limit` slice; only episodes that will actually run are pre-claimed. Added `test_debug_limit_orphans_do_not_block_resume` which would have failed before.

https://github.com/The-AI-Alliance/cube-harness/blob/9869c2931dde8233cd19035fdac094064d8483ac/src/cube_harness/exp_runner.py#L437-L443

2. **`save_failure` / XRay regression** — left for a separate PR since it requires wiring `status.json` error fields into `xray_utils.py` and is a self-contained change.

3. **Private storage methods across module boundary** — added `archive_episode(trajectory_id)` to the `Storage` Protocol and `FileStorage`; `_pre_claim` now calls that instead of reaching into `_episode_dir`/`_archive_episode`. Spec updated accordingly.

https://github.com/The-AI-Alliance/cube-harness/blob/9869c2931dde8233cd19035fdac094064d8483ac/src/cube_harness/storage.py#L36-L50

4. **Dead `is_retriable()`** — removed; `_should_relaunch` already has the same logic.

amanjaiswal73892 (approved):
Great PR!  Extensive tests are added and found issues are addressed. 
Follow-up: Create a issue for wiring xray to the statuses post merge.

PR #316, opened 2026-04-27
fix(exp_runner): correct task_id access path on EpisodeConfig

amanjaiswal73892:
## Summary
- Fixes `AttributeError: 'EpisodeConfig' object has no attribute 'task_id'` introduced in 0bbe3d71
- `EpisodeConfig` has no top-level `task_id` — the correct path is `episode.config.task_config.task_id`; fixed all four occurrences in `_run_sequentially_impl` and `_poll_ray`
- Adds a `test_run_sequentially` integration test so this regression is caught by CI going forward
- Adds `/// script` dependency block to `recipes/osworld/eval_azure_osworld.py`

## Test Plan
- [x] `pytest tests/test_experiment.py tests/test_episode.py` — 35 passed
- [x] New `test_run_sequentially` test exercises the exact code path that was broken

### DIFF
--- a/recipes/osworld/eval_azure_osworld.py
+++ b/recipes/osworld/eval_azure_osworld.py
@@ -1,3 +1,17 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "cube-harness",
+#     "osworld-cube",
+#     "cube-infra-azure",
+# ]
+#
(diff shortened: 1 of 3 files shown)


NicolasAG (approved):
Thanks for catching this!

PR #322, opened 2026-04-28
chore(web-cubes): migrate miniwob, webarena-verified, workarena to cube-standard 0.1.0rc7

NicolasAG:
Migration plan for the **web** group of cubes onto cube-standard `0.1.0rc7`,
which lands the merged [PR #118 (BenchmarkConfig split + CompositeBenchmark)](https://github.com/The-AI-Alliance/cube-standard/pull/118)
and [PR #121 (typed `TaskExecutionInfo` + `extra_info` removal)](https://github.com/The-AI-Alliance/cube-standard/pull/121).

This is **PR 1 of 3** for the cube-standard rc7 migration:

- **PR1 (this) — web cubes**: `miniwob`, `webarena-verified`, `workarena`
- PR2 — coding cubes: `swebench-live`, `swebench-verified`, `terminalbench` #323
- PR3 — `osworld` + `arithmetic` #324

The init commit on this branch only bumps `cube-standard>=0.1.0rc7` in the
three `pyproject.toml`s so CI / `uv.lock` already targets the new contract.
The real migration commits follow.

---

## Status

Draft. Init commit only — code migration commits land on this branch in follow-ups.

## Why these three together

None of the three carry heavy per-task data (`extra_info` is unused or already
migrated to typed `TaskMetadata` subclasses), so this PR is a pure
**layer-split** migration: it does not touch `TaskExecutionInfo` /
`load_task_execution_info()` / `cube install`. That keeps the diff small and
isolates the riskier coding/osworld migrations into PR2 and PR3.

### From PR #118 (BenchmarkConfig split)

`Benchmark` is split into two classes that mirror the existing
`TaskConfig` / `Task` split:

```
BenchmarkConfig   →  make(infra)  →  Benchmark
  serializable                       not serializable
  subset_from_list()                 _setup() / close()
  named_subset()                     spawn()
  install()                          _runtime_context
```

- `BenchmarkConfig` is the Pydantic, JSON-serializable layer that holds
  ClassVars (`benchmark_metadata`, `task_metadata`, `task_config_class`,
  `benchmark_class`) plus user-tunable instance fields (`task_ids`,
  `resources`, `tool_config`, `seed_generator`, `container_backend`).
- `Benchmark` is a plain Python class — not Pydantic — instantiated only by
  `BenchmarkConfig.make(infra)`. It owns `_setup()` / `close()` and writes
  shared infra references into `self._runtime_context`.
- `default_tool_config` was renamed to `tool_config`
  ([cube-standard#123](https://github.com/The-AI-Alliance/cube-standard/pull/123)).
- Subsets are now `task_ids: list[str] | None` on the config — no more
  `model_copy(deep=True)` + `__pydantic_private__` reset hacks.
- `TaskConfig.task_id` is a derived `@property`
  (`"{sub_bench_name}/{metadata.id}"` for composites, else `metadata.id`).
  Callers must emit `metadata=tm` instead of `task_id=tm.id`.
- `TaskConfig.metadata` is stamped onto each emitted config by
  `BenchmarkConfig.get_task_configs()`. `make()` reads `self.metadata`
  directly — no more `XxxBenchmark.task_metadata[self.task_id]` lookup,
  which removes the circular-import dodge in every cube's `task.py`.
- `AbstractSeedGenerator` is now a `TypedBaseModel` (was a plain
  `BaseModel`). Custom seed generators must round-trip with the `_type`
  discriminator — relevant to **workarena** below.

### From PR #121 (typed `TaskExecutionInfo`)

`TaskMetadata.extra_info` and `BenchmarkMetadata.extra_info` are **deleted**.
None of the web cubes used these as bags — `MiniWobTaskMetadata`,
`WebArenaVerifiedTaskMetadata`, and `WorkArenaTaskMetadata` are already typed
subclasses with named fields, so this PR has nothing to migrate on the
`TaskExecutionInfo` / `cube install` axis.

---

### `miniwob` ([cubes/miniwob](../tree/chore/cube-standard-rc7-web/cubes/miniwob))

- **Split.** `MiniWobBenchmark(Benchmark)` → `MiniWobBenchmarkConfig(BenchmarkConfig)`
  + new `MiniWobBenchmark(Benchmark)` (plain runtime class).
- **Move runtime state.** `_server_process`, `_stdout_file`, `_stderr_file`
  off the Pydantic class and onto the runtime class. The
  `model_config = {"arbitrary_types_allowed": True}` on the old class goes
  away — runtime state no longer collides with serialization.
- **Move `_setup` / `close`** (server start/teardown) to the runtime class.
  Publish `base_url` via `self._runtime_context["base_url"] = self.base_url`.
- **Rename** `default_tool_config` → `tool_config` on the config.
- **`get_task_configs()`** emits `MiniWobTaskConfig(metadata=tm, tool_config=…, base_url=…, …)`
  — drop `task_id=tm.id`, add `metadata=tm`.
- **`MiniWobTaskConfig.make()`** reads `self.metadata` directly — drop the
  `from miniwob_cube.benchmark import MiniWobBenchmark` circular-import dodge
  and the `MiniWobBenchmark.task_metadata[self.task_id]` lookup.
- **`debug.py.get_debug_benchmark(infra=None)`** signature change per the
  `cube.testing` contract; subset call becomes
  `MiniWobBenchmarkConfig(tool_config=PlaywrightConfig(...)).subset_from_list(_DEBUG_TASK_IDS, benchmark_name_suffix="debug").make(infra)`.
- **Add `MiniWobBenchmarkConfig.benchmark_class = MiniWobBenchmark`** ClassVar.
- **Update entry point** `pyproject.toml`:
  `miniwob-cube = "miniwob_cube.benchmark:MiniWobBenchmarkConfig"`.

### `webarena-verified` ([cubes/webarena-verified](../tree/chore/cube-standard-rc7-web/cubes/webarena-verified))

- **Split.** `WebArenaVerifiedBenchmark` → `WebArenaVerifiedBenchmarkConfig`
  + runtime `WebArenaVerifiedBenchmark`.
- **Move runtime state.** `_handle: ResourceHandle` (`PrivateAttr`) → runtime
  class. `_launch_and_configure()` and `_configure_wav_from_handle()` move
  onto the runtime class.
- **Drop `infra: InfraConfig | None` field on the config.** Per the rc7
  contract, infra is never stored on `BenchmarkConfig`; it flows through
  `make(infra)`. Provisioning is handled automatically by `make(infra)` for
  any `DockerServiceConfig` declared in `resources`. The runtime's `_setup()`
  then calls `self._launch_and_configure(infra)` to launch + populate
  `wav_config.environments`. Decision (per design): infra is passed by
  overriding `BenchmarkConfig.make(infra)` on this cube to forward it into
  the runtime constructor (`WebArenaVerifiedBenchmark(config, infra=infra)`)
  — kept as a per-cube override.
- **Rename** `default_tool_config` → `tool_config` (the
  `ToolboxConfig(tool_configs=[HarPlaywrightConfig(), SubmitResponseConfig()])`
  default stays).
- **`get_task_configs()`** emits `WebArenaVerifiedTaskConfig(metadata=tm, tool_config=…)`
  — `task_id=tm.id` removed.
- **`WebArenaVerifiedTaskConfig.make()`** reads `self.metadata` directly.
- **`debug.py.get_debug_benchmark(infra=None)`** signature change.
- **Add `benchmark_class` ClassVar** on the config.
- **Update entry point** to `WebArenaVerifiedBenchmarkConfig`.

### `workarena` ([cubes/workarena](../tree/chore/cube-standard-rc7-web/cubes/workarena))

- **Split.** `WorkArenaBenchmark` → `WorkArenaBenchmarkConfig` + runtime.
- **Move `_setup` / `close`** (no shared infra — both are no-ops; the
  log line moves to the runtime class).
- **`WorkArenaSeedGenerator(AbstractSeedGenerator)` upgrade.**
  `AbstractSeedGenerator` is now a `TypedBaseModel`. The class already
  inherits cleanly so the `_type` discriminator is auto-injected, but we
  need to confirm a round-trip serialization test passes (`config.model_dump_json()`
  → `WorkArenaBenchmarkConfig.model_validate_json(...)` produces an
  equivalent `seed_generator` with the right concrete subclass). Add a
  unit test if one isn't already there.
- **Move `_init_seed_generator` `model_validator`** to the config class
  (validators run at construction time on the Pydantic config, not the
  runtime). Drop the `object.__setattr__` (no longer needed — config is
  cleanly mutable during validation).
- **Rename** `default_tool_config` → `tool_config`.
- **`get_task_configs()`** emits `WorkArenaTaskConfig(metadata=tm, …)`,
  `seed=…` from the seed generator.
- **`WorkArenaTaskConfig.make()`** reads `self.metadata` directly — drop
  the circular-import dodge.
- **`debug.py.get_debug_benchmark(infra=None)`** signature change.
- **Add `benchmark_class` ClassVar** on the config.
- **Update entry point** to `WorkArenaBenchmarkConfig`.

---

## Out of scope for this PR

- `TaskExecutionInfo` / `cube install <bench>` plumbing — none of these
  three cubes need it.
- Per-task infra threading — none of these three cubes launch one container
  per task. SWE-bench, terminalbench, and osworld do; that pattern lands in
  PR2 and PR3.
- `cube-harness/` runner edits. Cubes only.

## Validation

Per cube:

- `make test` (where present) and `cube test <cube>` debug suite pass.
- Round-trip serialization of the config: `XxxBenchmarkConfig().model_dump_json()`
  → `XxxBenchmarkConfig.model_validate_json(...)`.
- For workarena: same round-trip preserves `seed_generator._type` and emits
  identical seed lists from the rehydrated config.


cubes/miniwob/pyproject.toml
```diff
@@ -19,6 +19,10 @@ build-backend = "uv_build"
 [tool.uv-build]
 include = ["src/miniwob_cube/task_metadata.json"]
 
+[tool.uv.sources]
+# TODO: drop once cube-standard 0.1.0rc7 ships to PyPI
+cube-standard = { git = "https://github.com/The-AI-Alliance/cube-standard", branch = "main" }
```
conte91: worth pinning to a tag or at least a commit instead of a branch?

cubes/workarena/src/workarena_cube/benchmark.py
```diff
@@ -92,34 +106,21 @@ class WorkArenaBenchmark(Benchmark):
...
+    benchmark_class: ClassVar[type[Benchmark]] = WorkArenaBenchmark
 
     meta_seed: int = 42
     n_seeds_l1: int = 10
     is_agent_curriculum: bool = True
 
     @model_validator(mode="after")
-    def _init_seed_generator(self) -> "WorkArenaBenchmark":
-        """Initialize seed_generator at construction time from benchmark fields."""
+    def _init_seed_generator(self) -> "WorkArenaBenchmarkConfig":
```
conte91: why quoted imports? you can just use `Self` (`from typing import Self`)

recipes/workarena.py
```diff
@@ -89,8 +89,10 @@ def main(debug: bool, agent: str, level: int) -> None:
         tools_configs.append(WorkArenaInfeasibleToolConfig())
     tool_config = ToolboxConfig(tool_configs=tools_configs)
 
-    # Configure WorkArena benchmark — filter to the requested level via named_subset
-    benchmark = WorkArenaBenchmark(default_tool_config=tool_config, n_seeds_l1=1).named_subset(f"l{level}")
+    # Configure WorkArena benchmark — filter to the requested level via named_subset,
+    # then materialize the runtime Benchmark via .make() (new BenchmarkConfig + Benchmark
+    # split from cube-standard 0.1.0rc7).
```
conte91: ? this comment seems very random

pyproject.toml
```diff
@@ -81,3 +81,7 @@ markers = [
 
 [tool.setuptools.packages.find]
 where = ["src"]
+
+[tool.uv.sources]
+# TODO: drop once cube-standard 0.1.0rc7 ships to PyPI
+cube-standard = { git = "https://github.com/The-AI-Alliance/cube-standard", branch = "main" }
```
conte91: [same as the above]

cubes/workarena/src/workarena_cube/debug.py
```diff
@@ -48,14 +48,13 @@ def make_debug_agent(task_id: str) -> CheatAgent:
...
         n_seeds_l1=1,
-        default_tool_config=WorkArenaCheatToolConfig(),
-    )
-    l1_bench = bench.named_subset("l1")
-    task_ids = list(l1_bench.task_metadata.keys())[:_DEBUG_N_TASKS]
-    return l1_bench.subset_from_list(task_ids)  # type: ignore
+        tool_config=WorkArenaCheatToolConfig(),
+    ).named_subset("l1")
+    task_ids = list(l1_cfg.tasks().keys())[:_DEBUG_N_TASKS]
+    return l1_cfg.subset_from_list(task_ids, benchmark_name_suffix="debug")  # type: ignore[return-value]
```
conte91: This [and all others] look dodgy. pytype is smart enough that `type: ignore` is usually a red flag that something is architecturally wrong. The `WebArenaVerifiedConfig` is using this same pattern to return the correct data type. Why wouldn't this class be able to do the same?

cubes/workarena/src/workarena_cube/benchmark.py
```diff
@@ -92,34 +106,21 @@ class WorkArenaBenchmark(Benchmark):
         },
         num_tasks=333,
     )
-    task_metadata: ClassVar[dict[str, WorkArenaTaskMetadata]]  # type: ignore - populated automatically at import time in Benchmark.__init_subclass__
+    task_metadata: ClassVar[dict[str, WorkArenaTaskMetadata]]  # type: ignore - populated automatically at import time
```
conte91: Comment makes little sense. I believe the `type` failure is about the fact that `task_metadata` is an invalid override here, because the parent class declares it as `dict[str, TaskMetadata]`. The bad news is that pytype is right to complain here. Overriding the type of `task_metadata` makes little sense w.r.t. inheritance. If I'm using a `BenchmarkConfig` object, then I have no information about which `BenchmarkConfig` I'm handling - so the child redeclaring as a "WorkArentTaskMetadata" makes little sense.

Because `Mapping`s are "idempotent" (This is a word i always forget, mighe be wrong 😆 ) this tactic of "leaving the generic type in the parent class and specializing in the child class" _would_ work if this was a function. e.g. `get_task_metadata() -> Mapping[str, TaskMetadata]` could correctly be overridden by `get_task_metadata() -> Mapping[str, WorkArenaTaskMetadata]`. The reason it doesn't work as-is is that if I have an object of the parent class, I am legally allowed to populate `task_metadata` (note it's a var - so it's read-write!) with an arbitrary `TaskMetadata` object, given the interface of `BenhcmarkConfig` - which is this warning is ultimately about :)

Apologies for the TED talk! Since you've been pushing hard on infra I want to make sure there is alignment on what I see as one of the main issue of the whole codebase (generic typing that gets specialized "incorrectly" thus removing all benefits that better type specialization could offer), so I cared about leaving a longer explanation of what the point is :)

cubes/workarena/src/workarena_cube/benchmark.py
```diff
@@ -55,13 +55,27 @@ def __call__(self, task_metadata: TaskMetadata) -> list[int]:
 
 
 class WorkArenaBenchmark(Benchmark):
-    """CUBE Benchmark for WorkArena ServiceNow tasks.
+    """Runtime pair — WorkArena tasks connect to a remote ServiceNow instance,
+    so there is no shared infrastructure to provision in _setup().
+    """
+
+    config: "WorkArenaBenchmarkConfig"  # type: ignore
```
conte91: why quoted imports? Can't we just move this class below the corresponding Config to remove forward declarations?

cubes/miniwob/src/miniwob_cube/benchmark.py
```diff
@@ -17,81 +17,54 @@
...
             stderr=self._stderr_file,
         )
-        startup_deadline = time.monotonic() + self.server_start_timeout
+        startup_deadline = time.monotonic() + cfg.server_start_timeout
         last_response_error: Exception | None = None
 
         while time.monotonic() < startup_deadline:
             if self._server_process.poll() is not None:
-                self._stderr_file.flush()
+                self._stderr_file.flush()  # type: ignore[attr-defined]
```
conte91: why are all these files defined as `object` and then we add `type: ignore` everywhere?

cubes/workarena/src/workarena_cube/benchmark.py
```diff
@@ -55,13 +55,27 @@ def __call__(self, task_metadata: TaskMetadata) -> list[int]:
...
+
+class WorkArenaBenchmarkConfig(BenchmarkConfig):
+    """CUBE BenchmarkConfig for WorkArena ServiceNow tasks.
 
     By default loads all task types from all levels (l1, l2, l3).
     Use named_subset() or subset_from_glob() in user-land to filter:
 
-        bench.named_subset("l1")                                   # L1 only
-        bench.named_subset("l2").subset_from_glob("in_human_curriculum", "True")  # L2 human curriculum
+        cfg.named_subset("l1")                                                  # L1 only
```
conte91, line 75: I wonder if the BenchmarkConfig class could have an easy way of discovering available parameters for `named_subset`; I believe since `Literal` is a subset of `str`, this class _should_ be able to override the parent's `named_subset` with:

```
def named_subset(self, subset: Literal['l1'] | Literal['l2]) -> ...
  return super().named_subset(self, subset)
```

this would make using this function in an IDE self-documenting since the allowed values are embedded in the type system :)

recipes/workarena.py
NicolasAG: I updated it

cubes/workarena/src/workarena_cube/benchmark.py
NicolasAG: we could, but then we would get another quoted type in the BenchmarkConfig here:
`benchmark_class: ClassVar[type[Benchmark]] = WorkArenaBenchmark`

cubes/workarena/src/workarena_cube/benchmark.py
NicolasAG: There is, and its `WorkArenaBenchmarkConfig.named_subsets()` :

```python
    @classmethod
    def named_subsets(cls) -> list[str]:
        """Return the names of all pre-defined subsets for this benchmark."""
        return list(cls.benchmark_metadata.named_subsets.keys())
```

Probably not worth it to override this for the following reasons:
1. Two sources of truth: `benchmark_metadata.named_subsets.keys()` is the source of truth, no need to add another
2. As mentioned above, discovery already exists: `WorkArenaBenchmarkConfig.named_subsets()`
3. Could be too restrictive for dynamic callers or users who introduce a new subset.

cubes/workarena/src/workarena_cube/debug.py
NicolasAG: this will be fixed once https://github.com/The-AI-Alliance/cube-standard/pull/124 lands.
For now, I will update the git paths in pyproject.toml to this branch

cubes/workarena/src/workarena_cube/benchmark.py
NicolasAG: I am removing the narrowed override; it was not used anyway.

conte91 (approved):
Self-approving after discussion with @NicolasAG

PR #327, opened 2026-04-29
chore: consolidate pre-commit hooks into .pre-commit-config.yaml

amanjaiswal73892:
## Summary

- Moves the pyproject local-path check and auto sign-off out of `.githooks/` and into `.pre-commit-config.yaml` as local hooks, matching how cube-standard manages hooks
- Unsets `core.hooksPath` so `pre-commit install` works without the \"Cowardly refusing\" error
- Shell scripts kept in `.githooks/` as entry points for the local hooks
- README updated with the one-time install command

## How to install after merging

```bash
pre-commit install --hook-type pre-commit --hook-type commit-msg --hook-type prepare-commit-msg
```


recursix (approved):
thanks

PR #345, opened 2026-05-01
fix(ray): upload editable cube installs as py_modules to Ray workers

recursix:
## Summary

- Ray creates a fresh worker venv from `pyproject.toml`/`uv.lock`, so editable installs (`cubes/*/src`, external packages like `cube_infra_toolkit`) are absent from workers and cause `ImportError` when Pydantic models from those packages are deserialized.
- Added `_editable_packages_for_ray()` which scans site-packages `.pth` files to discover editable package directories, then passes them as `runtime_env["py_modules"]` so Ray uploads them to every worker.
- The main `cube_harness` package is excluded — Ray already remaps its `.pth` file to the packed `working_dir`. Cube packages under `cubes/*/src` are included explicitly because they live inside `working_dir` but do NOT get `.pth` remapping.

## Root cause

Episodes submitted to Ray were stuck at `QUEUED` indefinitely. Workers failed silently during task deserialization with `ModuleNotFoundError: No module named 'swebench_verified_cube'` (and similar for other cube packages). The error surfaced only in Ray's internal logs — from the harness perspective, futures simply never resolved.

## Test plan

- [x] Ran 54-task SWE-bench Verified experiment; all episodes progressed from QUEUED → RUNNING with workers successfully importing cube packages
- [x] `_editable_packages_for_ray()` correctly excludes `cube_harness` and packages already inside `working_dir` that Ray handles natively
- [x] No `import site` inside function body (EX-001 compliant — `import site` at module level)

### DIFF
--- a/src/cube_harness/exp_runner.py
+++ b/src/cube_harness/exp_runner.py
@@ -1,7 +1,9 @@
 """Run experiments with Ray or sequentially."""
 
 import logging
+import site
 import time
+from pathlib import Path
 from uuid import uuid4
 
 import ray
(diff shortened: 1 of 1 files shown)


recursix:
Closing in favor of a documentation/launch-contract fix.

Root cause analysis: the  was not caused by Ray worker venv isolation — it was caused by running  from a Claude worktree. When  doesn't match uv's expected project path, uv ignores the project venv and creates an ephemeral environment at `~/.cache/uv/environments-v2/<hash>/`. That ephemeral env's .pth files point to the worktree path, which gets deleted after iteration, breaking imports for Ray workers.

The correct fix is at the launch level: always use `.venv/bin/python recipe.py` or `uv run --active recipe.py` (never bare `uv run` when VIRTUAL_ENV may be set to a different venv). This PR's approach of scanning .pth files and uploading packages as py_modules papers over the underlying bug and adds unnecessary complexity to the hot path.

PR #346, opened 2026-05-01
feat(xray): experiment search filter + agent/benchmark name resolution

recursix:
## Summary

- **Search filter**: multi-keyword search box beside the Refresh button. Space-separated terms are ANDed (Google-style), matched case-insensitively across experiment name, agent, model, and benchmark fields. Refresh respects the current search text.
- **Sort dropdown**: per-column sort beside the search filter (ascending/descending). avg_reward sorts numerically. Dropdown approach chosen because Gradio's built-in header sort requires `col_count="dynamic"`, which is incompatible with `static_columns` (needed to keep cells non-editable).
- **Agents tab `avg_step` column**: mean `n_env_steps` across finished trajectories (success/fail/max_steps), shown as "-" when none have finished.
- **Agent/benchmark name resolution** (from #356): resolves `agent_name` property and `benchmark_metadata` ClassVar directly from configs, so the Experiments table shows meaningful names even when the agent subclass module isn't importable. Also accepts the legacy `'benchmark'` key in `parse_exp_config`.

## Test plan

- [ ] Type a single term in the search box — table filters live on each keystroke
- [ ] Type two space-separated terms — only rows matching both are shown
- [ ] Click Refresh with a search term active — results stay filtered
- [ ] Clear the search box — full list is restored
- [ ] Change sort column — table reorders immediately; avg_reward sorts numerically
- [ ] Agents tab shows `avg_step` column with mean steps for finished runs
- [ ] Verify agent name and benchmark columns populate correctly for older experiments


amanjaiswal73892:
## Code Review

Reviewed against the [cube-harness constitution](.claude/rules/constitution.md) and [review rules](.claude/rules/review-rules.md).

### Summary

Solid PR overall — the `status.json` → XRay wiring is clean, the two-layer fallback for legacy traces is correct, and the test coverage (unit + fixture + Playwright e2e) is thorough. Two blocking issues were found and fixed directly on the branch in commit `2a83f7aa`. The remaining items are non-blocking.

---

### Fixed (committed to this branch)

**EX-002 — Duplicate injection logic in `load_missing_trajectory_stubs`** (`storage.py`)

The status-injection block was copy-pasted inline instead of calling `_maybe_inject_episode_status`. Had someone updated the method, the stub path would silently diverge. Replaced with a single call through a `stub_data` wrapper dict — all three load paths now share one implementation.

**TC-001 — `_traj_row_ids` race in `on_select_trajectory`** (`xray.py`)

`on_select_trajectory` was reading `state._traj_row_ids[row]` — a shared snapshot that `on_bg_load_tick` (firing every second during a live experiment) can overwrite between the user click and handler execution. A user clicking row N at the wrong moment could select the wrong trajectory. Fixed by recomputing traj_ids from `state.trajectories` at call time, which is always consistent with what `on_bg_load_tick` computes. As a side effect, the second redundant `build_trajectory_table` call in the same handler was eliminated.

---

### Non-blocking (follow-up welcome)

**EX-001 — `_render_retries` accesses `storage._episode_dir()` directly**
Private method access across the abstraction boundary. Won't break today but will silently fail if `FileStorage` internals change. Consider guarding with `isinstance(storage, FileStorage)` or exposing `load_retry_history` via the `Storage` protocol.

**SR-002 — `get_experiments_table_rows` is O(n_episodes) on every page load**
`_build_exp_status_cell` opens one `status.json` per episode directory. Fine for small experiments; worth an mtime-based cache or a `# TODO` comment for larger runs (500+ episodes).

**CC-001 — Silent `except Exception: pass` in `_build_exp_status_cell` and `_parse_experiment_config`**
Wrong status data surfaces with no trace. `logger.debug(exc)` would make these diagnosable without changing behaviour.

**CC-005 — `TERMINAL_OUTCOME_STATUSES` defined twice**
`TERMINAL_OUTCOME_STATUSES` (public frozenset, ~line 708) and `_TERMINAL_OUTCOME_STATUSES` (private, ~line 758) hold the same values. `_build_status_cell` uses the private one. Consolidate to one.

**Minor suggestions**
- `_map_episode_status`: use `if mapped is not None` instead of `or` to avoid a falsy-string edge case in future.
- `build_trajectory_table`: determine `include_seed` before the loop rather than post-hoc `del r["seed"]`.

amanjaiswal73892 (approved):

recursix:
Thanks for the review @amanjaiswal73892 — verified your `eeab1f64` fixes for EX-002 and TC-001 look correct (the `setdefault("metadata", {})` in `_maybe_inject_episode_status` returns the existing dict reference, so the in-place update propagates back to the caller's `metadata`).

Pushed `634bb7f1` addressing the non-blocking items:

- **EX-001** — `_render_retries` now guards with `isinstance(storage, FileStorage)` and returns a clear message for other backends.
- **CC-001** — silent `except Exception: pass` blocks in `_parse_experiment_config`, `_compute_exp_row`, and `get_experiments_table_rows` now log via `logger.debug(...)`.
- **CC-005** — dropped the duplicate `_TERMINAL_OUTCOME_STATUSES` private frozenset; everything uses the public `TERMINAL_OUTCOME_STATUSES` now.
- **`_map_episode_status`** — explicit `mapped is not None` instead of `or`.
- **`build_trajectory_table`** — `include_seed` precomputed once before the loop; rows only get the `seed` key when needed (no more post-hoc `del`).

**SR-002** (O(n_episodes) per page load) was addressed earlier in `c9923b73` — `.xray_summary.json` per-experiment cache, mtime-invalidated, only written once every episode is terminal (which now includes ghost-promoted STALE episodes from crashed coordinators).

PR #348, opened 2026-05-01
feat(swebench-verified): BashOnlySWEBenchTool, str_replace, field_validator, loop detection

recursix:
## Summary

- **`task.py`**: Add `@field_validator("fail_to_pass", "pass_to_pass", mode="before")` — the execution cache stores those fields as JSON-encoded strings; the validator transparently parses them, fixing a Pydantic validation error that caused all episodes to silently fail on workers.
- **`tool.py`**: Add `BashOnlySWEBenchTool` (bash-only variant for baseline agents), `str_replace` action (atomic targeted edits), `view` action (file contents with line numbers), and base64-chunked writes to avoid shell `exec` ARG_MAX limits.
- **`benchmark.py` / `debug.py` / `__init__.py`**: Migrate to the `Benchmark` / `BenchmarkConfig` split from cube-standard; `install()` / `uninstall()` use `cls.task_config_class.task_execution_cache_dir()` (fixes cache-dir mismatch — cube-standard issue #126). Export new public types.
- **`genny.py`**: Loop detection — track repeated identical `(action, args)` pairs and inject a `[LOOP]` warning into the next observation so the model self-corrects.
- **`swe_agent_recipe.py`**: Switch default tool to `BashOnlySWEBenchTool`, add `oracle_mode` flag, tighten system prompt with explicit file-editing workflow and quit guidance.

## Note on overlap with PR #344

Both this PR and #344 migrate `swebench_verified_cube/benchmark.py` to the new `Benchmark`/`BenchmarkConfig` API. Our version additionally incorporates the `task_config_class.task_execution_cache_dir()` fix (cube-standard issue #126) and the `BashOnlySWEBenchTool` exports. Whichever merges second will need a trivial rebase on `benchmark.py`.

## Test plan

- [x] Ran 54-task SWE-bench Verified experiment end-to-end; workers successfully imported cube packages and executed tasks (35.7% resolved after retry)
- [x] `field_validator` fix validated: `fail_to_pass` and `pass_to_pass` parse correctly from JSON-string cache entries
- [x] `str_replace` and `view` actions confirmed working in live agent trajectories
- [x] Loop detection fires correctly on repeated identical actions


cubes/swebench-verified-cube/src/swebench_verified_cube/task.py
```diff
@@ -71,6 +73,16 @@ class SWEBenchVerifiedExecutionInfo(TaskExecutionInfo):
     eval_timeout: int = 1800
     """Wall-clock seconds allowed for the evaluation test commands."""
 
+    # SWE-bench HuggingFace data ships these as JSON-encoded strings in some
+    # versions of the execution cache (e.g. '["tests/test_foo.py::test_bar"]').
+    # Normalise to list[str] regardless of which cache format was written.
+    @field_validator("fail_to_pass", "pass_to_pass", mode="before")
```
recursix: **Why this exists**

The SWE-bench HuggingFace dataset is inconsistent across versions: `fail_to_pass` and `pass_to_pass` arrive as a real `list[str]` in newer cache entries but as a JSON-encoded string (e.g. `'["tests/test_foo.py::test_bar"]'`) in older ones. Without this validator, Pydantic raises a validation error on old cache files. The `mode='before'` normalises both shapes to `list[str]` before the field type is applied.

cubes/swebench-verified-cube/src/swebench_verified_cube/task.py
```diff
@@ -96,7 +108,20 @@ def _exec(self) -> SWEBenchVerifiedExecutionInfo:
...
+        """Ensure /testbed source files are writable, then build the tool.
+
+        SWE-bench containers often have root-owned .py files (644) inside a
+        world-writable /testbed.  cp+mv re-owns each non-writable file in place:
+        mv unlinks via the parent directory (writable) and recreates with runtime
+        user ownership.  Running before relocate_if_readonly keeps the path stable
+        so conda editable installs (which point to /testbed) continue to work.
+        """
+        self._container.exec(
+            f"find {self.tool_config.working_dir} -not -path '*/.git/*' -name '*.py' ! -writable"
```
recursix, line 140: **Why this exists**

SWE-bench Docker images are built with `root` as the owning user and some `.py` files come out `644` (no write bit for the runtime user). `relocate_if_readonly` handles a fully read-only mount, but in the common case the *directory* is writable while individual files are not — so the agent's edits silently fail.

The `cp + mv` pattern re-owns each non-writable file in place without changing its path: `mv` unlinks via the parent directory (which is writable) and recreates the file under the runtime user. This runs before `relocate_if_readonly` so that conda editable installs pointing to `/testbed` remain valid.

cubes/swebench-verified-cube/src/swebench_verified_cube/tool.py
```diff
@@ -126,3 +126,67 @@ def write_file(self, path: str, content: str) -> str:
...
+    working_dir: str = "/testbed"
+    max_output_bytes: int = MAX_OUTPUT_BYTES
+
+    def make(self, container: Container | None = None) -> "BashOnlySWEBenchTool":
+        if container is None:
+            raise ValueError("BashOnlySWEBenchTool requires a container")
+        return BashOnlySWEBenchTool(config=self, container=container)
+
+
+class BashOnlySWEBenchTool(Tool):
```
recursix, line 143: **Why this exists**

The full `SWEBenchTool` exposes `bash`, `read_file`, `write_file`, and `str_replace` — a rich surface that adds parsing overhead and injects lint feedback into the observation stream.

`BashOnlySWEBenchTool` exposes only `bash`, matching the surface used by mini-SWE-agent (which scores ~74% on SWE-bench Verified). Agents edit files via inline Python, heredoc, or sed inside bash calls. The simpler surface reduces tool-call parsing failures and keeps observations clean.

`bash_unlimited` exists as a private escape hatch used by the task's `evaluate()` and `_apply_patch()` — it skips the output size cap so test output is never silently truncated during scoring.

src/cube_harness/agents/genny.py
```diff
@@ -500,11 +501,17 @@ def _choose_context(self) -> list[dict | Message]:
...
+        The step counter [Step X/N] is prepended to the final user message only — all
+        preceding messages are byte-for-byte identical across steps, preserving cache hits
+        on the system prompt, goal, hints, and summaries prefix.
         """
         messages = self._build_base_prompt(exclude_last_summary=self.config.enable_summarize)
         if self.config.enable_summarize and self.summaries:
             messages.append({"role": "assistant", "content": self.summaries[-1]})
         final_prompt = self.config.act_prompt if self.config.enable_summarize else self.config.react_prompt
+        if self.config.max_actions is not None:
+            final_prompt = f"[Step {self._actions_cnt + 1}/{self.config.max_actions}]\n\n{final_prompt}"
```
recursix, line 514: **Why this exists**

Without a step counter the agent has no signal for how much budget it has left. Agents tend to over-explore early and rush fixes at the end — labeling each turn `[Step X/N]` lets the model adjust pacing (e.g. skip exhaustive grep and commit to a best-guess fix when past the halfway point).

The counter is prepended to the *final* user message only (the act prompt / react prompt). All preceding messages — system prompt, goal, hints, summaries — are byte-for-byte identical across steps, so the prompt cache hit covers the expensive prefix and only the last message is re-encoded each turn.

cubes/swebench-verified-cube/src/swebench_verified_cube/task.py
NicolasAG: why not deleting your old execution_task_info cache folder?? the new one will be populated with the appropriate format. I don't think it is our responsibility to handle all dataset versions on huggingface

NicolasAG:
> * **`task.py`**: Add `@field_validator("fail_to_pass", "pass_to_pass", mode="before")` — the execution cache stores those fields as JSON-encoded strings; the validator transparently parses them, fixing a Pydantic validation error that caused all episodes to silently fail on workers.

I don't think it is our responsibility to handle all dataset versions on huggingface. 
Instead, when a dataset changes, users should delete their old execution_task_info cache folder (or call `BenchmarkConfig.uninstall()` and a new cache folder will be populated in their next run with `BenchmarkConfig.install()`

cubes/swebench-verified-cube/src/swebench_verified_cube/task.py
recursix: Good question. The issue is that `install()` is idempotent — it skips immediately if the cache directory already exists:

```python
if exec_cache_dir.exists() and any(exec_cache_dir.iterdir()):
    logger.info("Execution cache already populated, skipping installation")
    return
```

So upgrading the code does **not** refresh old caches. A user who installed before the `isinstance` fix landed in `_build_execution_info` (PR #344) will still have JSON-encoded strings on disk, and `install()` will never overwrite them.

The `field_validator` is 3 lines and makes the upgrade seamless — without it, those users would silently get a Pydantic validation error when a task runs on a worker. They would have to know to manually call `uninstall()` + `install()` to fix it, which also means re-downloading the full HuggingFace dataset.

Happy to drop it if you prefer to enforce an explicit reinstall — just means we need a migration note in the release.

recursix: Good point — dropped it. CUBE is not in production so we do not need backward compat with old cache formats; a reinstall is sufficient.

PR #352, opened 2026-05-02
chore(waa-cube): update cube-standard sources to main branch

amanjaiswal73892:
fix/waa-guest-agent-methods has been merged to cube-standard main. Drop the pinned rev in favour of tracking main directly.


kushasareen (approved):
All good

PR #361, opened 2026-05-05
fix(tbench): harden tool I/O, drop premature final_step, wire cube_uv assets mount

recursix:
## Summary

Iterating on the TerminalBench 40-task medium subset surfaced four properly-scoped harness bugs and one architectural extension. All proper fixes — no hacks. Stacks on `feat/terminalbench-run1` (existing branch) and depends on cube-standard #142 for the `assets_data` field used by the cluster-B fast path.

Depends-on: cube-standard/fix/recoverable-tool-arg-errors-and-assets-mount

## What's in here

* **`bash` hardening** in `TerminalBenchTool` (mirrors `BashOnlySWEBenchTool` from PR #348): ms→s timeout normalisation, `_MAX_BASH_TIMEOUT=900s`, head+tail truncation when `max_output_bytes` is exceeded (was dropping the tail), and `bash_unlimited` for the evaluator's own test.sh invocation.
* **`read_file` truncation** — was unbounded. An agent calling `read_file` on a 1.7MB g-code file blew the LLM context to 890K tokens and hung the worker on `ContentPolicyViolationError`. Now applies the same head+tail cap as `bash`.
* **`write_file` error reporting** — was always returning `"Wrote N bytes to ..."` even when the underlying `_exec` failed (e.g. write to read-only `/app`). Now checks exit codes and surfaces the error so the agent doesn't proceed with a phantom file.
* **Workdir rewrite for the agent prompt** in `TerminalBenchTask.reset`: when `relocate_if_readonly` moves CWD to `/tmp/app`, rewrite `/app/...` references in the task instruction the same way we already rewrite test scripts at `task.py:160-173`. Fixes ~5 tasks where the agent followed literal `/app/` wording while the evaluator looked under `/tmp/app/`.
* **Drop `final_step` instruction injection.** Empirically (run 5 → 6 on the iter subset): with `final_step` exposed, gpt-5.4-mini submitted prematurely after 2-5 actions on 7/9 short failures. Removing it (let auto-eval fire at step/cost budget) gained `constraints-scheduling`, `sqlite-db-truncate`, `distribution-search` as new wins. Removed the `_TASK_INSTRUCTIONS` constant, the `filter_actions` STOP_ACTION override, and the `append_submission_instructions` config flag.
* **`/opt/cube-assets/uv` fast path** in `_ensure_uv_preinstalled` (consumes the new `assets_data` slot from cube-standard #142): when the harness mounts a Linux `uv` binary, copy + chmod into `$HOME/.local/bin/` and create the env file. Bypasses the python3-bootstrap fallback that fails on minimal images (regex-log, extract-elf, merge-diff-arc-agi-task) lacking python3, curl, AND apt sources. In run 8 this fast path activated on **39/39 episodes** — every previously-broken evaluator now collects and runs pytest cleanly.
* **CLI surface**: `--sidecar-data` on `cube test terminalbench-cube` and `--assets-data` on the new iter recipe.
* **`recipes/genny2_terminalbench_iter_recipe.py`** — fixed 40-task subset chosen proportionally across the 14 categories, used as the iteration target so improvements are directly comparable across runs.

## Run progression on the 40-task subset (gpt-5.4-mini)

| Run | Wins | Key change |
|---|---|---|
| 2 (baseline, broken bash) | 5/40 | — |
| 3b (with workdir bug) | 2/40 | added `final_step` (regression) |
| 4 (workdir fix) | 6/38 | `/app`→`/tmp/app` rewrite |
| 5 (no `final_step` instr) | 6/40 | + read_file cap |
| 6 (no `final_step` action) | 7/38 | + cube-std arg validation, removed `filter_actions` |
| 8 (cluster B unblocked) | 6/39 | + EAI data mount of `uv` binary |

The cluster-B fast path didn't add new model wins on this run (the agent's solutions for those tasks are still wrong) but it unlocks ~5 tasks for any future model strong enough to solve them — the evaluator now runs to completion on every image.

## Test plan

- [x] `make lint` clean
- [x] cube-standard `pytest tests/test_tool.py` — 39/39 passing (4 new for the validation paths in dependency PR #142)
- [x] Live run: 8 iterative TerminalBench runs against `yul101`, ~$60 in compute. 0 episodes dying from `StepError` arg-error in run 8 (cube-std fix); cluster-B fast path activated on 39/39.
- [ ] Reviewer: run `cube test terminalbench-cube --toolkit --eai-profile yul101 --sidecar-data snow.allac.cube_sidecar` for the debug suite


recursix:
Re-opening as a focused PR off main with only the tool.py+task.py fixes — the previous branch was forked from feat/terminalbench-run1 and so the diff included unrelated prior commits (SWE-bench refactors, xray fixes, etc.). The debug.py CLI flag and iter recipe depend on the unmerged feat/terminalbench-run1 branch and will land with it.

PR #362, opened 2026-05-05
fix(tbench): harden tool I/O, drop self-defeating final_step injection

recursix:
Depends-on: cube-standard/dev
Depends-on: cube-standard/fix/recoverable-tool-arg-errors-and-assets-mount

## Summary

TerminalBench harness fixes + full migration to `TerminalTool`/`TerminalToolConfig` from cube-standard. No cube-specific tool subclass remains.

**cube-standard deps:**
- `dev` — `TerminalTool`/`TerminalToolConfig` base (PR #143, merged)
- `fix/recoverable-tool-arg-errors-and-assets-mount` — `assets_data` field for the `_ensure_uv_preinstalled` fast path (PR #142)

## What changed in the tool layer

`TerminalBenchTool` subclass deleted. Its responsibilities are now in better homes:

- **ms→s timeout normalization** → dropped. It was a workaround for smelly LLM behaviour; `max_timeout=900` already caps the worst case.
- **upload_file / upload_directory** → private `_upload_file` / `_upload_directory` methods on `TerminalBenchTask` (task-setup helpers belong on the task)

Default config is now `TerminalToolConfig(working_dir="/app", max_timeout=900, enable_file_actions=True)`.

## What changed in the task layer

- `_working_dir` cached in `_build_tool()` — no more `self.tool._config.working_dir` reads (eliminated `# type: ignore[attr-defined]` comments)
- `isinstance` guards changed from `TerminalBenchTool` → `TerminalTool`; one guard at the top of each public/private method that calls `self.tool.bash()` / `self.tool.write_file()`
- `reset()`, `evaluate()`, `_ensure_uv_preinstalled()`, `_install_python3_nonroot()` all properly guarded

## Earlier fixes (existing commits)

- `TerminalBenchTask`: `/app`→`/tmp/app` path rewrite in instruction text; drop `final_step` action injection
- Fast-path for `ToolkitInfraConfig.assets_data` (`_ensure_uv_preinstalled`): bypasses python3-bootstrap on minimal images

## Run progression (gpt-5.4-mini, 40-task subset)

| Run | Wins | Key change |
|---|---|---|
| 2 (baseline) | 5/40 | — |
| 4 (workdir fix) | 6/38 | `/app`→`/tmp/app` rewrite |
| 6 (no final_step) | 7/38 | removed filter_actions |
| 8 (assets fast path) | 6/39 | uv binary mount |

## Test plan

- [x] `make lint` clean
- [x] Live: 8 iterative TerminalBench runs against `yul101`


recursix (commented):
Per-block review notes — what each change fixes and why it lives where it does. Happy to split into smaller commits if any of these stand on their own.

cubes/terminalbench-cube/src/terminalbench_cube/task.py
```diff
@@ -120,7 +120,17 @@ def reset(self) -> tuple[Observation, dict[str, Any]]:
...
-        return Observation.from_text(self._exec.instruction), {
+        instruction = self._exec.instruction
+        # Mirror the test-script rewrite: when /app is read-only and we relocated
+        # the working dir to /tmp/app, the agent's instructions still reference
+        # /app verbatim (from terminal-bench task.yaml).  Rewrite so the agent
+        # writes to the same path the evaluator checks.
+        assert isinstance(self.tool, TerminalBenchTool)
+        app_dir = self.tool._config.working_dir  # type: ignore[attr-defined]
+        if app_dir != "/app":
+            instruction = re.sub(r"/app(?=[\s/.,;:)\]'\"]|$)", app_dir, instruction)
```
recursix: **`/app` → `/tmp/app` rewrite in the agent prompt**

*Bug it fixes.* When the toolkit user can't write `/app` (the common case — non-root containers), `_build_tool` already calls `relocate_if_readonly` to copy `/app` to `/tmp/app` and points the tool at the new dir. The evaluator side already rewrites `/app/...` references in test scripts at `task.py:160-173`. But the **agent's prompt** still says verbatim things like *“Create file /app/server.py”* (from `terminal-bench`'s task.yaml). The agent obediently writes to `/app/...`, gets a phantom success (see the `write_file` change), and the evaluator under `/tmp/app` sees nothing. We saw this kill 5+ tasks on the iter subset (kv-store-grpc, portfolio-optimization, large-scale-text-editing, build-cython-ext, modernize-scientific-stack).

*Why here.* This is the same kind of rewrite already happening in `evaluate()` for test scripts — same input (instruction text mentioning `/app`), same target (`app_dir`), same regex shape. Putting it in `reset()` keeps the rewrite at the only spot where the instruction is first surfaced to the agent. Done as `re.sub` with a lookahead instead of `str.replace` so we don't clobber `/application` or `/appx`.

*If we don't do this.* The agent silently writes to the wrong directory, the evaluator silently scores 0, and we waste the entire step budget.

cubes/terminalbench-cube/src/terminalbench_cube/task.py
```diff
@@ -234,6 +244,28 @@ def _ensure_uv_preinstalled(self) -> None:
...
+            logger.info("Using mounted /opt/cube-assets/uv for uv install")
+            self.tool.bash(
+                "export HOME=/tmp/fakehome && "
+                "mkdir -p $HOME/.local/bin && "
+                "cp /opt/cube-assets/uv /opt/cube-assets/uvx $HOME/.local/bin/ && "
+                "chmod +x $HOME/.local/bin/uv $HOME/.local/bin/uvx && "
+                "printf 'export PATH=\"$HOME/.local/bin:$PATH\"\\n' > $HOME/.local/bin/env",
+                timeout=30,
+            )
+            return
```
recursix, line 314: **Cube-assets fast path for `uv` install**

*Bug it fixes.* The existing fallback chain in this method (apt root → apt-get-download non-root → bootstrap pip → install uv) cascades to a logged warning and returns when none of those work. That happens on minimal task images that ship without python3 / curl / apt sources at all (e.g. `regex-log`, `extract-elf`, `merge-diff-arc-agi-task`). Result: the env file `/tmp/fakehome/.local/bin/env` is never created, test.sh fails on `source $HOME/.local/bin/env: No such file or directory`, pytest never runs, eval reports `passed=0/total=0`. The judge tagged 5 tasks on our subset as exactly this `env_failure`.

*Why here.* This is the right hand-off point: cube-standard PR #142 adds `ToolkitInfraConfig.assets_data` that mounts a Linux `uv` binary read-only at `/opt/cube-assets/`. The fast path runs before the python3-bootstrap chain so we skip 4-5 bash calls + a network/package fetch when the mount is present. If the mount isn't configured, the existing fallback runs unchanged.

*Two non-obvious bits.*
- `test -f` not `test -x` — EAI data mounts strip the execute bit (mode 0600), so the source binary is *not* executable. Confirmed empirically with a probe job (`ls -la /opt/cube-assets/`).
- `cp` + `chmod +x` after copying out — the mount is RO so we can't chmod the source. Copying to `$HOME/.local/bin/` first lets us mark the copy executable.

*If we don't do this.* All 5 cluster-B tasks stay broken at the evaluator level even when the model produces a correct solution. With the fix, run 8 saw the fast path activate on **39/39 episodes**, every previously-broken evaluator now collects and runs pytest cleanly.

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
             parts.append(result.stdout)
         if result.stderr:
             parts.append(result.stderr)
         if result.exit_code == 124:
-            parts.append(f"[error] Command timed out after {timeout}s")
+            parts.append(f"[timed out after {timeout}s]")
         elif result.exit_code != 0:
             parts.append(f"[exit_code: {result.exit_code}]")
-        output = "\n".join(parts) if parts else "(no output)"
+        return "\n".join(parts) if parts else "(no output)"
```
recursix: **Extract `_run_bash` helper + `_MAX_BASH_TIMEOUT` constant**

*Why here.* Pure refactor in service of the next two changes — the bash core needed to be reused in two places (the @tool_action public surface with truncation/timeout-cap, and `bash_unlimited` for the evaluator's own internal use). The TODO note flags that this whole block is a near-duplicate of `BashOnlySWEBenchTool` from PR #348; we want to extract them to a shared `cube-tools` package eventually but it's out of scope for this PR.

*Behaviour change.* The timeout error string changed from `"[error] Command timed out after Ns"` to `"[timed out after Ns]"` — same content, more compact, matches the format the SWE-bench tool already uses (so traces look uniform across cubes).

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
+            timeout = timeout // 1000
+        timeout = min(timeout, self._MAX_BASH_TIMEOUT)
+        output = self._run_bash(command, timeout=timeout)
         encoded = output.encode("utf-8")
         if len(encoded) <= self._config.max_output_bytes:
             return output
-        return encoded[: self._config.max_output_bytes].decode("utf-8", errors="ignore") + "\n[truncated]"
+        head = encoded[:5000].decode("utf-8", errors="ignore")
+        tail = encoded[-5000:].decode("utf-8", errors="ignore")
+        return f"{head}\n[... {len(encoded) - 10000} bytes elided ...]\n{tail}"
```
recursix: **`bash`: ms→s timeout normalisation, `_MAX_BASH_TIMEOUT` clamp, head+tail truncation**

*Three bugs in one method.*

1. **ms confusion.** Multiple LLMs (gpt-5.4-mini, claude) hallucinate `timeout=120000` thinking the unit is milliseconds (matches the JS Date / many SDKs). Without normalisation the container exec runs for 120000 seconds = ~33 hours, blocking the worker until a higher-level timeout fires. Fix: any `timeout > 3600` is treated as ms and divided by 1000.
2. **No upper bound.** Without `_MAX_BASH_TIMEOUT`, an LLM could legitimately ask for a 10-hour bash run, hanging a worker. Hard cap at 900s (matches the evaluator's own `max_test_timeout_sec`).
3. **Tail-dropping truncation.** The old `encoded[:max] + '\n[truncated]'` behaviour drops the *last* lines of long output. For most agent commands the *tail* is what carries the error message / final result; cutting it off makes the output less actionable. Head+tail is what `BashOnlySWEBenchTool` does and what GitHub Actions / many CI runners do for the same reason.

*Why here.* All three are properties of the public `bash` tool action (the agent-facing entry point), not of the underlying `_run_bash` helper. Internal callers (`bash_unlimited`, the test-harness invocation) bypass these intentionally.

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
+
+    def bash_unlimited(self, command: str, timeout: int = 120) -> str:
+        """Unlimited bash for internal use (evaluate, test harness)."""
+        return self._run_bash(command, timeout=timeout)
```
recursix: **`bash_unlimited` — internal escape hatch.**

*Why here.* The `evaluate()` method has to run `test.sh` with the task's own `max_test_timeout_sec` (sometimes 900s, sometimes longer) and wants the *full* pytest output even if it's huge — truncating in the middle would garble the parsed results. So the evaluator calls `bash_unlimited` and gets the raw `_run_bash` behaviour. Public `bash` retains all its caps.

*Naming follows the convention used in `BashOnlySWEBenchTool`.*

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
         result = self._exec(f"cat {shlex.quote(path)}")
         if result.exit_code != 0:
             return f"Error reading {path}: {result.stderr or result.stdout}"
-        return result.stdout
+        encoded = result.stdout.encode("utf-8")
+        if len(encoded) <= self._config.max_output_bytes:
+            return result.stdout
+        head = encoded[:5000].decode("utf-8", errors="ignore")
+        tail = encoded[-5000:].decode("utf-8", errors="ignore")
+        return f"{head}\n[... {len(encoded) - 10000} bytes elided ...]\n{tail}"
```
recursix: **`read_file` truncation**

*Bug it fixes.* `read_file` had no size cap. On `gcode-to-text` an agent called `read_file('/app/text.gcode')` on a 1.7 MB G-code file; the entire content got returned, the LLM context request totalled **890,033 tokens** (limit: 272,000), Azure rejected with `ContentPolicyViolationError`, and the worker hung in a retry loop until I killed it. (Same root cause that makes `bash` apply the cap.)

*Why here.* The cap belongs at the `@tool_action` boundary — the underlying `_exec("cat …")` should faithfully read whatever's on disk; it's the agent-facing surface that needs to bound what gets fed back into the LLM context.

*Symmetry.* Same head/tail format as `bash` so agents see consistent truncation markers across both tools.

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
         """Write content to a file in the sandbox."""
-        self._exec(f"mkdir -p {shlex.quote(str(Path(path).parent))}")
+        mkdir = self._exec(f"mkdir -p {shlex.quote(str(Path(path).parent))}")
+        if mkdir.exit_code != 0:
+            return f"Error creating parent dir for {path}: {mkdir.stderr or mkdir.stdout}"
         escaped = content.replace("'", "'\\''")
-        self._exec(f"printf '%s' '{escaped}' > {shlex.quote(path)}")
+        write = self._exec(f"printf '%s' '{escaped}' > {shlex.quote(path)}")
+        if write.exit_code != 0:
+            return f"Error writing {path}: {write.stderr or write.stdout}"
```
recursix: **`write_file`: stop lying about success**

*Bug it fixes.* The old method ran `mkdir -p` and `printf ... > path`, ignored both exit codes, and *unconditionally* returned `"Wrote N bytes to {path}"`. So if `path` was in a read-only directory (most commonly `/app` on toolkit; see the `task.py` rewrite), the bytes never landed but the agent got a happy success message and proceeded to invoke the file — only to get "file not found" downstream. Worst kind of bug: silent data loss with a confident success report.

*Why here.* `_exec` already returns the exit code; we just have to stop discarding it. Returning the stderr/stdout in the error message gives the agent enough information to retry against a writable path (which it does — saw recovery attempts in run 4 traces).

*Why not at `_exec` level.* `_exec` is also called for read-only operations (`cat`, `test -f`) where non-zero exit is normal data, not failure. The error-as-Observation policy is per-action.

cubes/terminalbench-cube/src/terminalbench_cube/task.py
```diff
@@ -58,6 +58,10 @@ class TerminalBenchTask(Task[TerminalBenchTaskMetadata]):
     accept_agent_stop: bool = True
     oracle_mode: bool = False
 
+    @property
+    def action_set(self) -> list:
+        return super().action_set + [STOP_ACTION]
```
recursix: **Expose STOP_ACTION in the action schemas**

*Bug it fixes.* STOP_ACTION (schema name: final_step) is defined in cube-standard's cube.task and handled by Task.step() when accept_agent_stop=True. But it was never added to the *exposed* action schemas -- task.action_set calls filter_actions(tool.action_set), and TerminalBenchTool has no @tool_action for final_step. Meanwhile the system prompt and goal template both tell the agent to 'call final_step when done'. With tool_choice=required the agent must emit a tool call; final_step is not in the schema so it falls back to bash and echoes the string, burning 70-100 steps in a loop.

*Run-9 evidence.* Judge assigned primary_blame=agent_scaffolding to 12 of 72 failures, all with the same pattern. Several dozen more had it as a secondary factor. Representative: hf-model-inference had a working Flask server at step 20 then spent 77 steps echoing the string.

*Why here.* TerminalBenchTask.action_set is the single place that controls what schemas the harness serialises into the LLM tool-call list. Overriding it to append STOP_ACTION exposes it without touching the tool or the harness loop.

*Precedent.* legacy_generic_agent.py already appends STOP_ACTION to its own action set for the same reason.

cubes/terminalbench-cube/src/terminalbench_cube/task.py
```diff
@@ -95,6 +99,12 @@ def _build_tool(self) -> None:
     def reset(self) -> tuple[Observation, dict[str, Any]]:
         self.tool.reset()
 
+        # Fast-fail if the container sandbox is broken (all bash commands return
+        # exit_code=1 with no output). Avoids wasting 100 agent steps on a dead env.
+        health = self.tool.bash("echo ok", timeout=15)
+        if "ok" not in health:
+            raise RuntimeError(f"Container health check failed for {self.metadata.id!r}: {health!r}")
```
recursix, line 106: **Container health check**

*Bug it fixes.* prove-plus-comm_ep63: the container started but every bash command -- including 'echo hi' and 'true' -- returned exit_code=1 with empty output. The agent correctly diagnosed the broken environment but was forced to spam no-op tool calls for 100 steps (tool_choice=required). The health check catches this at reset time and raises RuntimeError, which the episode runner records as an infrastructure error rather than wasting the full step budget.

*Why here.* reset() is the handoff point between infra setup and agent start. A broken container can't be recovered during the episode, so fast-fail here is strictly better than slow-fail at step 100.

*What it costs.* One 'echo ok' per episode before the task archive is extracted -- negligible compared to the archive extraction + uv install that follows.

recursix (commented):
Two additional task-layer fixes from run-9 judge analysis.

cubes/terminalbench-cube/src/terminalbench_cube/task.py
```diff
@@ -95,6 +95,12 @@ def _build_tool(self) -> None:
     def reset(self) -> tuple[Observation, dict[str, Any]]:
         self.tool.reset()
 
+        # Fast-fail if the container sandbox is broken (all bash commands return
+        # exit_code=1 with no output). Avoids wasting 100 agent steps on a dead env.
+        health = self.tool.bash("echo ok", timeout=15)
```
recursix, line 104: Test comment

recursix, line 104: Empirical motivation: `prove-plus-comm ep63` in the overnight 89-task run had a fully broken container where every `bash` call returned `exit_code=1` with no output. The agent wasted all 100 steps attempting to work around what looked like permission errors. One extra `bash echo ok` at reset (15s timeout) is negligible compared to 100 wasted steps and a container GC charge.

cubes/terminalbench-cube/src/terminalbench_cube/task.py
```diff
@@ -120,7 +126,17 @@ def reset(self) -> tuple[Observation, dict[str, Any]]:
...
             self._ensure_uv_preinstalled()
 
-        return Observation.from_text(self._exec.instruction), {
+        instruction = self._exec.instruction
+        # Mirror the test-script rewrite: when /app is read-only and we relocated
+        # the working dir to /tmp/app, the agent's instructions still reference
+        # /app verbatim (from terminal-bench task.yaml).  Rewrite so the agent
+        # writes to the same path the evaluator checks.
+        assert isinstance(self.tool, TerminalBenchTool)
+        app_dir = self.tool._config.working_dir  # type: ignore[attr-defined]
```
recursix: Pairs with the test-script path rewrite at line ~158. `relocate_if_readonly` moves `working_dir` from `/app` to `/tmp/app` when the runtime user cannot write `/app` (common on EAI Toolkit images). Without this, the agent follows the literal task instructions ("modify /app/src/..."), writes to the read-only path, while the evaluator looks in `/tmp/app`. The result: the agent reports success, pytest fails, reward=0.

cubes/terminalbench-cube/src/terminalbench_cube/task.py
```diff
@@ -234,6 +250,28 @@ def _ensure_uv_preinstalled(self) -> None:
...
             return
 
+        # Fast path: cube_assets EAI data mount (see ToolkitInfraConfig.assets_data).
+        # When the harness mounts /opt/cube-assets/ with a uv binary, copy it
+        # directly — bypasses the python3-bootstrap path that fails on minimal
+        # images lacking python3, curl, AND apt sources.  Note: EAI data mounts
+        # are read-only and strip the execute bit (mode 0600), so we use ``-f``
+        # for the probe and ``chmod +x`` after copying into the writable HOME.
+        assets_probe = self.tool.bash(
+            "test -f /opt/cube-assets/uv && test -f /opt/cube-assets/uvx && echo YES || echo NO",
```
recursix, line 301: Fast path bypassing the python3 → uvx install chain, which fails on minimal task images (LaTeX, pure-C toolchains) lacking python3, curl, AND apt sources. EAI data mounts strip the execute bit (mode 0600), so we probe with `test -f` not `test -x`, then `chmod +x` after copying to writable HOME. This unblocked `regex-log`, `extract-elf`, `compile-compcert` which were silently scoring `passed=0/total=0`.

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
+    # bash() impl mirrors BashOnlySWEBenchTool (feat/swebench-verified-improvements).
+    # TODO: extract to a shared cube-tools package.
 
-    @tool_action
-    def bash(self, command: str, timeout: int = 120) -> str:
-        """Execute a bash command in the sandbox and return its output."""
-        result = self._exec(command, timeout=timeout)
+    _MAX_BASH_TIMEOUT: int = 900
+
+    def _run_bash(self, command: str, timeout: int = 120) -> str:
```
recursix: Extracted from `bash()` so `bash_unlimited()` (line 84) can serve the evaluator's internal `test.sh` invocations without head+tail truncation. Without this separation, pytest output for long test suites could be silently cut off and the reward would be computed on incomplete results.

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
+
+    @tool_action
+    def bash(self, command: str, timeout: int = 120) -> str:
+        """Execute a bash command in the sandbox and return its output.
+
+        Args:
+            command: Shell command. cwd=/app.
+            timeout: Seconds (NOT milliseconds), max 900.
+        """
+        if timeout > 3600:  # agent sent ms instead of seconds (e.g. 120000)
```
recursix: LLMs sometimes emit `timeout=120000` (milliseconds — OpenAI default is 120s expressed as ms). The heuristic (> 3600 → divide by 1000) silently corrects this. Without it, the container exec receives a 120 000-second timeout and the worker hangs for ~33 hours or until Ray kills it.

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
+            timeout = timeout // 1000
+        timeout = min(timeout, self._MAX_BASH_TIMEOUT)
+        output = self._run_bash(command, timeout=timeout)
         encoded = output.encode("utf-8")
         if len(encoded) <= self._config.max_output_bytes:
             return output
-        return encoded[: self._config.max_output_bytes].decode("utf-8", errors="ignore") + "\n[truncated]"
+        head = encoded[:5000].decode("utf-8", errors="ignore")
```
recursix: Head+tail truncation instead of head-only. Critical for long build logs and compiler output where the actual error is near the end (linker errors, failing test names). The `[... X bytes elided ...]` marker also gives the agent a size signal so it knows how much is missing and can ask for specific sections via `read_file`.

cubes/terminalbench-cube/src/terminalbench_cube/tool.py
```diff
@@ -44,40 +44,70 @@ def _exec(self, command: str, **kwargs: Any) -> ExecResult:
...
+        tail = encoded[-5000:].decode("utf-8", errors="ignore")
+        return f"{head}\n[... {len(encoded) - 10000} bytes elided ...]\n{tail}"
 
     @tool_action
     def write_file(self, path: str, content: str) -> str:
         """Write content to a file in the sandbox."""
-        self._exec(f"mkdir -p {shlex.quote(str(Path(path).parent))}")
+        mkdir = self._exec(f"mkdir -p {shlex.quote(str(Path(path).parent))}")
+        if mkdir.exit_code != 0:
+            return f"Error creating parent dir for {path}: {mkdir.stderr or mkdir.stdout}"
```
recursix: Previously `write_file` on a read-only `/app` silently returned `"Wrote N bytes to /app/..."`. The agent would then spend 30–50 steps trying to execute a file that was never written. Surfacing the error immediately lets the agent retry with a writable path (`/tmp/app/...` or `/tmp/`).

cubes/terminalbench-cube/tests/test_tool_io.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def _fail(stderr: str = "error") -> ExecResult:
+    return ExecResult(stdout="", stderr=stderr, exit_code=1, duration_seconds=0.0)
+
+
+# ── bash() timeout normalisation ─────────────────────────────────────────────
+
+
+def test_bash_millisecond_timeout_converted_to_seconds() -> None:
```
recursix: **Bug this test pins: LLM-emitted ms timeouts caused 33-hour container exec hangs**

**Symptom**: Overnight Ray runs would accumulate hung workers that never returned. Tracing showed the container `exec()` calls were being dispatched with `timeout=120000` (seconds), causing them to wait up to 33 hours. The root cause: the LLM — particularly GPT-4 variants following OpenAI's tooling conventions — would emit `timeout=120000`, interpreting the parameter as milliseconds (the OpenAI SDK default unit). `bash()` passed this value directly to `container.exec(timeout=...)`, which expects seconds.

**Fix**: values >3600 are assumed to be milliseconds and divided by 1000 before passing to `container.exec`.

**Certainty**: High. The threshold (>3600) is conservative: no legitimate second-based timeout for a single bash call should exceed an hour.

cubes/terminalbench-cube/tests/test_tool_io.py
```diff
@@ -0,0 +1,129 @@
...
+    tool, container = _make_tool()
+    container.exec.return_value = _ok("ok")
+
+    tool.bash("echo ok", timeout=120_000)
+
+    _, kwargs = container.exec.call_args
+    assert kwargs["timeout"] == 120
+
+
+def test_bash_timeout_capped_at_900s() -> None:
```
recursix: **Why `timeout=1800`, not `timeout=9000`**

The ms-conversion check fires first: any value >3600 is divided by 1000. So `9000 → 9s`, which is well under the 900s cap — the cap would never be tested. `1800` is ≤3600, so no ms-conversion happens, and since 1800 > 900 the cap fires correctly. The test name documents the intent; this comment exists because the "obvious" choice (a large number like 9000) silently tests the wrong code path.

cubes/terminalbench-cube/tests/test_tool_io.py
```diff
@@ -0,0 +1,129 @@
...
+    """Output under max_output_bytes is returned as-is."""
+    tool, container = _make_tool(max_output_bytes=1000)
+    container.exec.return_value = _ok("hello")
+
+    result = tool.bash("echo hello")
+
+    assert result == "hello"
+
+
+def test_bash_large_output_uses_head_tail() -> None:
```
recursix: **Bug this test pins: head-only truncation hid compiler errors from the agent and blew context limits**

**Symptom 1 (context overflow)**: An agent called `read_file` on a 1.7 MB G-code file during a build task. The previous truncation returned the first N bytes only. The result ballooned the LLM context to ~890K tokens (model limit: 272K), causing a `ContentPolicyViolationError` that hung the worker.

**Symptom 2 (missed errors)**: On build-log tasks (e.g., `make` output), compiler errors and linker failures always appear at the end of the output. Head-only truncation meant the agent never saw the actionable error message — it only saw the initial build steps succeeding, leading to 30-50 confused retry steps.

**Fix**: head+tail truncation with an elision marker in the middle. The tail is the same size as the head (both at `max_output_bytes // 2`).

**The key assertion here**: `assert "BBBB" in result` — the tail ("BBBB" bytes) must be present. This is the regression check that head-only truncation would fail.

cubes/terminalbench-cube/tests/test_tool_io.py
```diff
@@ -0,0 +1,129 @@
...
+
+    assert "AAAA" in result  # head preserved
+    assert "BBBB" in result  # tail preserved — the key regression check
+    assert "bytes elided" in result
+
+
+# ── write_file() error propagation ───────────────────────────────────────────
+
+
+def test_write_file_error_on_mkdir_failure() -> None:
```
recursix: **Bug these tests pin: `write_file` silently reported success on read-only paths**

**Symptom**: The agent would write to `/app/out.py` (read-only in the TerminalBench container), receive `"Wrote 42 bytes to /app/out.py"`, then spend 30-50 steps confused about why running `/app/out.py` failed or why the file seemed empty. The silent failure was found by reviewing XRay trajectories where the agent looped on `python /app/out.py` without ever changing its write target.

**Fix**: `write_file` now checks the exit code of both the `mkdir -p` and the `python3 -c` write step, and returns an explicit error string on failure.

**What the two tests cover**: mkdir failure (e.g., read-only parent directory) and write failure (e.g., mkdir succeeded but the file itself is read-only). Both must produce an error string; neither should return `"Wrote N bytes"`.

Issue #332, opened 2026-04-30
sweep_stale_statuses: RUNNING episodes not swept if kill-and-retry within step_timeout window


recursix:
## Problem

If you kill a running experiment (e.g. `kill <pid>`) and immediately relaunch with `--retry`, any episodes that were `RUNNING` at kill time will **not** be picked up by the retry run.

`sweep_stale_statuses()` only marks a `RUNNING` entry as `STALE` if its `last_heartbeat_at` is older than `step_timeout_s + cancel_grace_s` (default: 1800 + 120 = 1920s). A freshly killed worker has a recent heartbeat, so the sweep leaves it in `RUNNING`. Since `RUNNING` is not in `RETRIABLE_STATUSES`, `_should_relaunch()` skips it — the episode is silently lost.

## Expected behaviour

Killing a process (SIGTERM/SIGKILL) should cause orphaned `RUNNING` statuses to become retriable on the next run, regardless of heartbeat age. The worker is provably dead.

## Possible fix

On `run_with_ray` / `run_sequentially` startup, perform an **unconditional** RUNNING → STALE sweep (ignoring the heartbeat threshold) for any episodes whose `last_heartbeat_at` predates the current process start time. This is safe because no workers from the current process have written heartbeats yet at that point.

Alternatively, expose a `--force-sweep` flag to the retry recipe that overrides the threshold.

## Workaround

Wait > 32 minutes after killing before relaunching with `--retry`, so the heartbeat threshold is exceeded and the sweep catches the orphaned entries.

PR #365, opened 2026-05-07
fix(harness): retry/queue robustness — config archive, XRay sweeps, experiment heartbeat

recursix:
## Summary

Comprehensive robustness pass on the harness retry / queue / status machinery, surfaced by overnight SWE-bench Verified and TerminalBench runs. Each fix below was independently rediscovered across feature branches; this PR consolidates them onto `main` so other threads can drop their local copies.

### 1. `_archive_episode` was destroying the planning record

`Experiment.prepare_episodes` writes `episode_config.json` exactly once and the retry pipeline relies on it to discover retriable work. The old `_archive_episode` renamed the whole directory, hiding the config inside `.archived_<ts>/` so subsequent retries silently dropped the task.

**Fix:** read `episode_config.json` before rename and rewrite into the fresh `ep_dir`. Trajectory data, status, logs all stay archived; only the planning record is preserved.

### 2. XRay viewer was killing in-flight tasks; `ray.get` could hang forever

- `_promote_ghost_episodes` was promoting `QUEUED` episodes to STALE on every UI refresh. In a 89-task TerminalBench batch (`--n-parallel 10`) the last tasks legitimately sit QUEUED for hours — opening the viewer mid-run silently killed 3 unstarted tasks.
- `ray.get(task_ref)` had no timeout, so a worker hanging during teardown froze the entire `_poll_ray` loop.

**Fix:** XRay only promotes `RUNNING` episodes (later refined — see #4); `ray.get` now takes `timeout=step_timeout_s + cancel_grace_s`.

### 3. Phase-aware setup timeout + kill-and-retry RUNNING sweep (#332)

Setup hangs (container boot, env reset) had to wait the full `step_timeout_s` (30 min default) before being killed. And a re-launched run could not sweep RUNNING episodes whose heartbeat predated the new process start, even though those workers were provably dead.

**Fix:**
- New `setup_timeout_s` (default `4 × step_timeout_s = 7200s`; container scheduling can queue) applied when `current_step == 0` in `_kill_stale_workers`. Setup hangs detected on their own budget without shortening agent-step budget.
- `sweep_stale_statuses` accepts `process_start_s` and force-sweeps RUNNING episodes whose heartbeat predates it — fixes the kill-and-retry race where a freshly killed worker has a recent heartbeat.
- Episode heartbeat writes are now fail-open (don't abort the run on transient I/O errors).

### 4. Experiment-level heartbeat + XRay dead-driver detection

`experiment_status.json` (new file at `output_dir/` root) is written by the runner driver, not by Ray workers. XRay reads it to decide whether the driver is alive, and only then sweeps orphan QUEUED → STALE.

- `run_with_ray`: heartbeats every 30 s from `_poll_ray`.
- `run_sequentially`: heartbeats before each `episode.run()` call (driver is the worker, so heartbeat may lag by one episode wall-clock).
- `_driver_alive` is mode-aware: ray trusts the experiment heartbeat; sequential falls back to per-episode heartbeats to avoid false positives during long episodes.
- `_promote_ghost_episodes` now also promotes QUEUED → STALE when the driver is confirmed dead, and **immediately** promotes RUNNING → STALE in sequential mode (driver == worker; process death kills both).
- Old experiments without `experiment_status.json` keep their prior behaviour (QUEUED preserved).

### 5. Centralised the RUNNING → STALE predicate

`should_sweep_running_to_stale` in `episode_status.py` is now the single source of truth for *is this RUNNING worker dead?* — used by both `sweep_stale_statuses` (runner-side) and `_promote_ghost_episodes` (viewer-side). Removes the silent duplication that would have rotted.

### 6. Cleanup pass on the new experiment-status code

- New `_experiment_lifecycle()` context manager wraps init + COMPLETED/INTERRUPTED terminal write. Shed ~25 lines of duplicated lifecycle code from each runner.
- New `ExperimentStatus.heartbeat()` method collapses the *set fields, try-write, swallow* pattern into one call at every heartbeat site.
- Added a final-counter flush at the end of `_poll_ray` and the sequential loop — without it, the last batch of completions (or any work shorter than `_EXP_HEARTBEAT_INTERVAL_S`) was lost from the on-disk record.

## Test plan

- [x] **Unit tests** — 13 cases in `tests/test_xray_promote_ghost.py` covering QUEUED-promotion gated by driver liveness (both modes), sequential fast-sweep on driver death, ray waiting for `GHOST_TIMEOUT` even when driver is dead, terminal episodes never re-promoted, backward compat for old experiments without `experiment_status.json`.
- [x] **Integration tests** — 4 new cases in `tests/test_retry_integration.py`:
  - `test_experiment_status_sequential_happy_path` — COMPLETED + correct counters
  - `test_experiment_status_sequential_interrupted` — INTERRUPTED on driver crash
  - `test_experiment_status_ray_happy_path` — Ray heartbeat lifecycle (slow, real Ray cluster)
  - `test_experiment_status_interrupted_promotes_queued` — full handoff: runner crashes mid-run → XRay's `_promote_ghost_episodes` sweeps orphan QUEUED → STALE.
- [x] Full local suite: 627 passing (pandas-dependent xray UI tests excluded as on main).
- [x] `make lint-check` clean.
- [x] Slow Ray integration tests pass locally (~45 s).

## Notes

- The original two-fix scope grew to a six-fix scope after Alex's review pass — the bugs were tightly entangled (config preservation, XRay sweeps, heartbeats, dead-driver detection all touch the same control plane). Reviewer mental model is the same: trust this PR is the canonical retry/queue cleanup.
- `ContainerBackend` shows up as a CI ruff failure on a file we don't touch (`cubes/swebench-verified-cube/src/swebench_verified_cube/task.py`). Minimal `# noqa: F401` patch included to unblock CI; full removal of `container_backend` across all cubes is tracked as a separate cleanup PR.

### DIFF
--- a/src/cube_harness/analyze/xray_utils.py
+++ b/src/cube_harness/analyze/xray_utils.py
@@ -26,6 +26,7 @@ from cube_harness.core import AgentOutput, Trajectory, TrajectoryStep
 from cube_harness.episode_status import STATUS_FILENAME, EpisodeStatus
 from cube_harness.episode_status import TERMINAL_STATUSES as _EPISODE_TERMINAL_STATUSES
 from cube_harness.exp_runner import DEFAULT_CANCEL_GRACE_S, DEFAULT_STEP_TIMEOUT_S
+from cube_harness.experiment_status import EXPERIMENT_STATUS_FILENAME, ExperimentStatus
 from cube_harness.llm import LLMCall
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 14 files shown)


src/cube_harness/storage.py
```diff
@@ -171,7 +171,16 @@ def save_trajectory(self, trajectory: Trajectory, allow_overwrite: bool = False)
...
     def _archive_episode(self, ep_dir: Path) -> None:
         archived = ep_dir.parent / f"{ep_dir.name}{ARCHIVED_MARKER}{time.time()}"
+        # Preserve episode_config.json so the retry pipeline still discovers this
+        # episode via _episode_config_dirs(). The config is written once at
+        # experiment prep time (Experiment.prepare_episodes) and is not re-saved
+        # on retry, so the archive must not destroy it.
+        config_src = ep_dir / "episode_config.json"
+        config_bytes = config_src.read_bytes() if config_src.exists() else None
         ep_dir.rename(archived)
+        if config_bytes is not None:
```
recursix, line 181: **What:** Before renaming `ep_dir` to its `.archived_<ts>` form, slurp `episode_config.json` into memory; after the rename, recreate `ep_dir` with just that file.

**Fix:** restores the retry pipeline. `_episode_config_dirs()` discovers retriable work by listing non-archived `episodes/*` dirs that contain `episode_config.json`. Without this preservation, archiving a terminal episode hides the task from every subsequent retry pass — the work is silently dropped.

**Why this layer (not the callers):** `_archive_episode` is a leaf called from three sites — `save_trajectory(allow_overwrite=True)` (line 157), `Episode._open_status` (episode.py:154), and the public `archive_episode(traj_id)` (line 182, used by `exp_runner.py:74`). Patching any one caller leaves the other two paths still destructive. Fixing at the leaf upholds the invariant for all of them.

**Why not restructure storage to keep config outside `ep_dir/`:** the layout's defining invariant is "everything for this episode lives in one directory," which is what makes archival a single atomic `rename()`. Splitting the config out would force every read/write site (`save_episode_config`, `load_episode_config`, `_episode_config_dirs`, the resume scan in `experiment.py`) to learn about a second location — much larger blast radius for the same end state.

**Why not re-save the config on retry instead:** the config is *plan* data (task identity, agent config, max_steps), written once at `Experiment.prepare_episodes`. Re-saving it during `_open_status` would imply the plan can change per attempt, which it can't — the retry pipeline relies on the plan being stable across attempts.

src/cube_harness/analyze/xray_utils.py
```diff
@@ -457,7 +464,7 @@ def _promote_ghost_episodes(exp_dir: Path) -> None:
         if not ep_dir.is_dir() or ".archived_" in ep_dir.name:
             continue
         status = EpisodeStatus.read(ep_dir / STATUS_FILENAME)
-        if status is None or status.status not in ("RUNNING", "QUEUED"):
+        if status is None or status.status != "RUNNING":
```
recursix: **What:** the viewer's ghost-promotion sweep now skips QUEUED episodes entirely. Previously it promoted both RUNNING and QUEUED.

**Fix:** in a large batch (89 tasks × `--n-parallel 10`), the tail of the queue legitimately sits QUEUED for many hours waiting for a Ray worker slot. Opening the viewer mid-run was promoting those un-dispatched tasks to STALE — silently killing 3 unstarted gpt-5.4 TerminalBench episodes in the run that prompted this fix.

**Why this layer:** the viewer is a *read-mostly observer*. Its only signal is heartbeats on disk; it has no view of the Ray scheduler. QUEUED means "no worker is touching this yet," so heartbeat-based timeouts are semantically wrong here — there's nothing to heartbeat. The runner's `sweep_stale_statuses()` (in `experiment.py`, called at resume-time and shutdown) does have scheduler context and is the authoritative cleaner for QUEUED orphans. Two cleaners with different views were stepping on each other.

**Why not give QUEUED episodes a heartbeat upstream:** that would be inventing fake state to satisfy the viewer's existing logic. The fix is to teach the viewer that the heartbeat model only applies to RUNNING.

**Why not raise the timeout instead of skipping:** the previous attempt already used a 1h orphan threshold and still fired on long queues. Any threshold short of "as long as the runner allows" is a tunable that will be wrong again next time the parallelism changes. Skipping is the correct invariant.

src/cube_harness/exp_runner.py
```diff
@@ -280,7 +280,7 @@ def _poll_ray(
         for task_ref in done:
             traj_id = ref_to_traj_id[task_ref]
             try:
-                traj: Trajectory = ray.get(task_ref)
+                traj: Trajectory = ray.get(task_ref, timeout=step_timeout_s + cancel_grace_s)
```
recursix, line 358: **What:** bound `ray.get(task_ref)` with the same `step_timeout_s + cancel_grace_s` budget the runner already uses for `_kill_stale_workers`.

**Fix:** if a worker hangs during teardown (observed concretely with `eai job kill` blocking on toolkit infra), `ray.get()` blocks forever. The whole `_poll_ray` loop is single-threaded, so once it stalls there, `_kill_stale_workers` no longer runs and the entire experiment freezes — no further dispatch, no shutdown.

**Why this layer:** Ray's `ray.get()` is by-design unbounded — Ray has no opinion about how long a task *should* take. The runner does: it already enforces the same budget out-of-band via `_kill_stale_workers`. Capping `ray.get` locally with the same value just makes the in-band path consistent with the kill path; without it, the kill path can't even fire because the loop never returns.

**Why not push this upstream into Ray:** Ray's scheduling model intentionally separates worker liveness from task-result waiting. Workers that are still doing useful compute shouldn't be killed by a `ray.get` timeout in the driver. We're not asking Ray to change that; we're applying our domain-specific budget at the one call site where the driver can't afford to block.

**Why not move this to the worker / `Episode`:** the worker can't know that the *driver* needs to keep polling. The deadlock is a driver-loop concern; the fix belongs in the driver loop.

recursix (commented):
Inline rationale for each block — what it does, what it fixes, and why it lives where it does (vs. moving the fix elsewhere).

PR #367, opened 2026-05-08
swebench-verified: fix _normalize_django_directive double-append

recursix:
## Bug

`_normalize_django_directive` could produce a double-appended method name for entries where the method is already embedded in class_path.

**Bug:** When a SWE-bench entry stores a directive as `test_foo (module.Class.test_foo)` — where the method name is already the last segment of class_path — the old code produced `module.Class.test_foo.test_foo`. Django's `runtests.py` rejects that path, returning a non-zero exit code even when the agent's patch was correct.

**Symptoms:** Judge analysis flagged ~5 episodes as `eval_brittle` with `primary_blame=eval_brittle` and `intervention=eval_fix`. All were Django repo tasks where the agent's patch was mechanically correct (gold-patch baseline reproduced the issue). The verifier returned reward=0 due to the double-appended test path.

**Fix:** Guard: if `class_path.endswith(f".{method}")`, return `class_path` directly without appending again.

**Verification:** 8 unit tests in `tests/test_django_normalize.py` cover the standard case, the already-normalized case, nested paths, and pass-through for unrecognized formats. Not independently verified via a controlled re-run of the flagged episodes; fix is by logic inspection.

> **Note:** This fix also exists on `feat/llm-cache-control` (PR #353) which touched `task.py`. This PR targets `main` directly; if #353 lands first, this can be closed (no conflict — different lines) or cherry-picked onto #353's base.

### DIFF
--- a/cubes/swebench-verified-cube/src/swebench_verified_cube/task.py
+++ b/cubes/swebench-verified-cube/src/swebench_verified_cube/task.py
@@ -241,6 +241,11 @@ class SWEBenchVerifiedTask(Task[SWEBenchVerifiedTaskMetadata]):
         m = re.match(r"^(\w+)\s+\(([^)]+)\)$", directive.strip())
         if m:
             method, class_path = m.group(1), m.group(2)
+            # Some SWE-bench entries already include the method name as the last
+            # segment of class_path (e.g. "test_foo (module.Class.test_foo)").
+            # Avoid duplicating it into "module.Class.test_foo.test_foo".
+            if class_path.endswith(f".{method}"):
+                return class_path
             return f"{class_path}.{method}"
         return directive  # already in the right format or unrecognised — pass through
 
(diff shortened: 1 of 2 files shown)


cubes/swebench-verified-cube/src/swebench_verified_cube/task.py
```diff
@@ -241,6 +241,11 @@ def _normalize_django_directive(directive: str) -> str:
         m = re.match(r"^(\w+)\s+\(([^)]+)\)$", directive.strip())
         if m:
             method, class_path = m.group(1), m.group(2)
+            # Some SWE-bench entries already include the method name as the last
+            # segment of class_path (e.g. "test_foo (module.Class.test_foo)").
+            # Avoid duplicating it into "module.Class.test_foo.test_foo".
+            if class_path.endswith(f".{method}"):
```
recursix, line 247: **Why this guard exists:** Some SWE-bench entries store the directive as `test_foo (module.Class.test_foo)` — the method name is already the last segment of `class_path`. Without this check, the code produced `module.Class.test_foo.test_foo`, which Django's `runtests.py` rejects. Judge analysis found ~5 episodes that got `reward=0` on a correct patch (`eval_brittle` label).

PR #368, opened 2026-05-08
swebench-verified: head+tail truncation for bash output

recursix:
Depends-on: cube-standard/dev

## Bug

Bash tool output truncation was head-only. Agents running long commands (test suites, large greps) had the actionable information — pytest summary, exit codes, error tracebacks — silently cut off.

**Bug:** `_truncate_output` returned `encoded[:max_output_bytes]` + `[truncated]`. For a typical pytest run where the failure summary appears at the bottom, the agent received only the header lines and had no information about what actually failed.

**Symptoms:** Not independently quantified. Observed pattern in trace analysis: agents re-ran identical long commands multiple times, consistent with not having received a usable error message. Pattern is common enough to be the subject of issue #363.

**Fix:** Replace head-only truncation with head(20%)+tail(80%). The 20/80 split is chosen because pytest `-rA` summary and non-zero exit messages are tail-positioned; the agent needs those to decide next steps.

**Verification:** Unit tests in `tests/test_terminal_tool.py` (cube-standard) verify boundary conditions, byte accounting, unicode safety, and the head/tail proportions. Not verified via controlled A/B benchmark re-run.

**Refs:** Companion fix for `swebench-live-cube` is in PR #369. Local bash tool deleted; migrated to `TerminalTool`/`TerminalToolConfig` from `cube.tools.terminal` (cube-standard/dev).


cubes/swebench-verified-cube/src/swebench_verified_cube/tool.py
```diff
@@ -70,7 +70,13 @@ def bash(self, command: str, timeout: int = 120) -> str:
         encoded = output.encode("utf-8")
         if len(encoded) <= self._config.max_output_bytes:
             return output
-        return encoded[: self._config.max_output_bytes].decode("utf-8", errors="ignore") + "\n[truncated]"
+        head = self._config.max_output_bytes // 5
```
recursix: **Why 20/80 split instead of 50/50:** Pytest `-rA` summary lines, non-zero exit codes, and error tracebacks are tail-positioned. With a 100kB cap this gives ~20kB of context (imports, collected items) and ~80kB of actionable output. The old head-only truncation gave the agent only the header with no information about what failed — traced to repeated identical command re-runs in loop analysis.

PR #369, opened 2026-05-08
swebench-live: head+tail truncation for bash output

recursix:
Depends-on: cube-standard/dev

## Bug

Same fix as companion PR #368 for `swebench-verified-cube`: bash tool output truncation was head-only, hiding actionable tail content (pytest summary, exit codes, error tracebacks).

**Bug:** `_truncate_output` returned only the first `max_output_bytes` bytes. For long test suite output, the agent saw only the beginning (imports, collected items) and not the failure summary at the bottom.

**Symptoms:** Not independently quantified. Observed pattern in trace analysis: agents re-ran identical long commands multiple times after truncated output, consistent with not seeing a usable result. See issue #363.

**Fix:** Replace head-only with head(20%)+tail(80%) truncation.

**Verification:** Unit tests in `tests/test_terminal_tool.py` (cube-standard) verify the logic. Not verified via controlled A/B benchmark re-run.

**Refs:** Companion fix for `swebench-verified-cube` is PR #368. Local bash tool deleted; migrated to `TerminalTool`/`TerminalToolConfig` from `cube.tools.terminal` (cube-standard/dev).


cubes/swebench-live-cube/src/swebench_live_cube/tool.py
```diff
@@ -63,7 +63,13 @@ def bash(self, command: str, timeout: int = 120) -> str:
         encoded = output.encode("utf-8")
         if len(encoded) <= self._config.max_output_bytes:
             return output
-        return encoded[: self._config.max_output_bytes].decode("utf-8", errors="ignore") + "\n[truncated]"
+        head = self._config.max_output_bytes // 5
```
recursix: **Why 20/80 split instead of 50/50:** Pytest `-rA` summary lines, non-zero exit codes, and error tracebacks are tail-positioned. With a 100kB cap this gives ~20kB of context (imports, collected items) and ~80kB of actionable output. The old head-only truncation gave the agent only the header with no information about what failed — traced to repeated identical command re-runs in loop analysis.

PR #370, opened 2026-05-08
llm: replace completion_with_retries with tenacity 120s-ceiling backoff

recursix:
## Bug

`litellm.completion_with_retries` hardcodes its exponential backoff to `wait_exponential(multiplier=1, max=10)` — a 10 s ceiling that's not configurable. Under heavy load (Anthropic `overloaded_error`, provider rate limits), a single failed request kills an entire multi-step episode with no chance of recovery, because all retries are exhausted in ~50 s while the provider window is often minutes long.

A 150-step episode that hit one transient 502 at step 73 would raise immediately and record `FAILED`, wasting all prior work.

## Fix

Own the retry loop with `tenacity.Retrying` directly — essentially the same 5 lines litellm wraps internally, but with the three parameters that matter actually exposed:

- `wait_exponential(multiplier=2, max=120)` — sleeps grow 2s → 4s → 8s → 16s → 32s → 64s → 120s (cap), outlasting typical overload windows.
- `retry_if_exception_type(...)` — only retriable errors are caught: `InternalServerError`, `ServiceUnavailableError`, `RateLimitError`, `Timeout`, `APIConnectionError`. Permanent errors (`NotFoundError`, `AuthenticationError`, etc.) surface immediately instead of wasting `num_retries` attempts.
- `reraise=True` — callers see the actual exception, not tenacity's `RetryError` wrapper.

`tenacity` was already a direct dependency, so no new packages.

## Verification

**Unit tests** (`TestRetryBackoff`, patch `tenacity.wait_exponential`):
- succeeds on first attempt → 1 call
- retries transient error then succeeds → 2 calls
- permanent error not retried → 1 call, raises immediately
- exhausted retries re-raises → `num_retries` calls then raises

**Integration tests** (`TestRetryIntegration`, real tenacity loop, only `time.sleep` patched):
- attempt count matches `num_retries`
- sleep durations are exponential (~2s, ~4s with `multiplier=2`)
- `num_retries=0` → 1 attempt (tenacity minimum)
- permanent error → 0 sleeps, 1 attempt
- successful retry returns the final response, not an artefact

## Follow-up

`LLMConfig.retry_strategy` is now dead config (was only consumed by `completion_with_retries`). Will be removed in a separate cleanup.

### DIFF
--- a/src/cube_harness/llm.py
+++ b/src/cube_harness/llm.py
@@ -3,11 +3,20 @@
 import pprint
 from datetime import datetime
 from functools import partial
-from typing import Callable, List, Literal
+from typing import Any, Callable, List, Literal
 from uuid import uuid4
 
+import litellm
+import tenacity
 from cube.core import TypedBaseModel
-from litellm import Message, completion_with_retries
+from litellm import Message
+from litellm.exceptions import (
+    APIConnectionError,
+    InternalServerError,
+    RateLimitError,
+    ServiceUnavailableError,
+    Timeout,
+)
 from litellm.utils import token_counter
 from pydantic import Field
 
@@ -80,8 +89,6 @@ class LLM:
             "model": self.config.model_name,
             "temperature": self.config.temperature,
             "max_completion_tokens": self.config.max_completion_tokens,
-            "num_retries": self.config.num_retries,
-            "retry_strategy": self.config.retry_strategy,
             "tool_choice": self.config.tool_choice,
             "parallel_tool_calls": self.config.parallel_tool_calls,
             "tools": prompt.tools,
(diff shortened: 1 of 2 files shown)


src/cube_harness/llm.py
```diff
@@ -184,10 +191,36 @@ def __call__(self, prompt: Prompt) -> LLMResponse:
...
         return LLMResponse(message=response.choices[0].message, usage=usage)
 
+    def _completion_with_retry(self, **kwargs: Any) -> Any:
+        """Call litellm.completion with exponential backoff on transient errors.
+
+        litellm's completion_with_retries caps its backoff at 10 s, which is too
+        short for Anthropic overloaded_error responses under heavy load. We own the
+        retry loop here to get a proper 120 s ceiling.
+        """
+        _RETRIABLE = (
```
recursix, line 111: **Why these specific error types and not all exceptions:** Only transient server-side errors are retried. `NotFoundError`, `AuthenticationError`, `BadRequestError` etc. are permanent — retrying them wastes worker slots with zero chance of success. PR #357 pins exactly this: a bad model name should fail on the first attempt, not after `num_retries` attempts.

src/cube_harness/llm.py
```diff
@@ -184,10 +191,36 @@ def __call__(self, prompt: Prompt) -> LLMResponse:
...
+            InternalServerError,
+            ServiceUnavailableError,
+            RateLimitError,
+            Timeout,
+            APIConnectionError,
+        )
+        retryer = tenacity.Retrying(
+            wait=tenacity.wait_exponential(multiplier=2, max=120),
```
recursix, line 119: **Why max=120s (not the default 10s):** `litellm.completion_with_retries` caps backoff at 10 s. Under Anthropic `overloaded_error` at high load, servers typically need 30–90 s to recover. With a 10 s cap, retries all fire before the server is ready, exhausting the budget on guaranteed failures. 120 s was chosen as a safe upper bound that covers Anthropic's documented recovery window without holding a Ray worker slot indefinitely.

PR #372, opened 2026-05-08
fix(runner): stop XRay from killing QUEUED tasks; add ray.get timeout

recursix:
## Summary

Two bugs caught during overnight 89-task TerminalBench runs:

1. **XRay silently killed QUEUED tasks** — `_promote_ghost_episodes()` (called on every XRay UI refresh) was promoting QUEUED episodes to STALE using a 1-hour orphan threshold. In a large batch (89 tasks, `--n-parallel 10`) the last tasks legitimately sit QUEUED for many hours before a worker slot opens. Opening the viewer mid-run silently killed 3 unstarted tasks in a gpt-5.4 run. Fix: only promote RUNNING episodes from the viewer. QUEUED orphan cleanup belongs exclusively to `sweep_stale_statuses()` in `experiment.py`, which runs at resume-time or runner shutdown.

2. **`ray.get()` had no timeout** — if a worker hangs during teardown (e.g. `eai job kill` blocks), the entire `_poll_ray` loop freezes and `_kill_stale_workers` never runs again. Fix: pass `timeout=step_timeout_s + cancel_grace_s` to `ray.get()`, matching the same budget already used for worker kill decisions.

## Test plan

- [ ] `make lint` clean
- [ ] Run XRay viewer on a live experiment with QUEUED episodes — confirm none are promoted to STALE
- [ ] Verify `_poll_ray` recovers after a hung worker (via `ray.get` timeout)

### DIFF
--- a/src/cube_harness/analyze/xray_utils.py
+++ b/src/cube_harness/analyze/xray_utils.py
@@ -444,10 +444,17 @@ _XRAY_CACHE_FILENAME = ".xray_summary.json"
 
 
 def _promote_ghost_episodes(exp_dir: Path) -> None:
-    """Write STALE into status.json for RUNNING/QUEUED episodes whose heartbeat is dead.
+    """Write STALE into status.json for RUNNING episodes whose heartbeat has gone silent.
 
     Fixes experiments where the runner crashed without marking episodes terminal.
     This lets the cache machinery treat the experiment as finished so it can be frozen.
+
+    Only RUNNING episodes are promoted — QUEUED episodes are legitimately waiting for
+    a Ray worker slot and must not be touched here.  In a large batch (e.g. 89 tasks,
+    --n-parallel 10) the last tasks can sit QUEUED for many hours; promoting them from
+    the viewer would silently kill tasks that the runner is about to dispatch.
+    QUEUED orphan cleanup is handled exclusively by sweep_stale_statuses() in
+    experiment.py, which runs either at resume-time or on runner shutdown.
     """
     episodes_dir = exp_dir / "episodes"
     if not episodes_dir.exists():
@@ -457,7 +464,7 @@ def _promote_ghost_episodes(exp_dir: Path) -> None:
         if not ep_dir.is_dir() or ".archived_" in ep_dir.name:
             continue
         status = EpisodeStatus.read(ep_dir / STATUS_FILENAME)
-        if status is None or status.status not in ("RUNNING", "QUEUED"):
+        if status is None or status.status != "RUNNING":
             continue
         hb = status.last_heartbeat_at or status.started_at
         if now - hb > GHOST_TIMEOUT:
--- a/src/cube_harness/exp_runner.py
+++ b/src/cube_harness/exp_runner.py
@@ -280,7 +280,7 @@ def _poll_ray(
         for task_ref in done:
             traj_id = ref_to_traj_id[task_ref]
             try:
-                traj: Trajectory = ray.get(task_ref)
+                traj: Trajectory = ray.get(task_ref, timeout=step_timeout_s + cancel_grace_s)
                 logger.info(
                     "Completed trajectory %s with %d steps (%d agent steps, %d environment steps)",
                     traj_id,
--- a/tests/test_xray_promote_ghost.py
+++ b/tests/test_xray_promote_ghost.py
@@ -1,0 +1,77 @@
+"""Unit tests for _promote_ghost_episodes (xray_utils).
+
+Covers the fix in PR #372: QUEUED episodes must never be promoted to STALE
+by the viewer's refresh path — only sweep_stale_statuses() (runner-owned)
+may touch them.
+"""
+
+from __future__ import annotations
+
+import time
+from pathlib import Path
+
+import pytest
+
+from cube_harness.analyze.xray_utils import _promote_ghost_episodes
+from cube_harness.episode_status import STATUS_FILENAME, EpisodeStatus
+
+
+def _write_status(ep_dir: Path, status: str, age_seconds: float) -> None:
+    ep_dir.mkdir(parents=True, exist_ok=True)
+    s = EpisodeStatus(
+        status=status,
+        task_id="test-task",
+        episode_id=0,
+        started_at=time.time() - age_seconds,
+        last_heartbeat_at=time.time() - age_seconds,
+    )
+    s.write(ep_dir / STATUS_FILENAME)
+
+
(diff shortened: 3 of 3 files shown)


src/cube_harness/analyze/xray_utils.py
```diff
@@ -444,10 +444,17 @@ def _parse_experiment_config(exp_dir: Path) -> dict[str, str]:
 
 
 def _promote_ghost_episodes(exp_dir: Path) -> None:
```
recursix, line 446: Updated docstring reflects the narrowed contract: this function only promotes RUNNING episodes (not QUEUED). The previous docstring described the 2-threshold approach that was introduced as an intermediate fix; this version is cleaner — QUEUED orphan cleanup belongs exclusively to `sweep_stale_statuses()` in `experiment.py`, which runs at resume-time or runner shutdown, not on every UI refresh tick.

src/cube_harness/analyze/xray_utils.py
```diff
@@ -457,7 +464,7 @@ def _promote_ghost_episodes(exp_dir: Path) -> None:
         if not ep_dir.is_dir() or ".archived_" in ep_dir.name:
             continue
         status = EpisodeStatus.read(ep_dir / STATUS_FILENAME)
-        if status is None or status.status not in ("RUNNING", "QUEUED"):
+        if status is None or status.status != "RUNNING":
```
recursix, line 467: The key fix: filter to `RUNNING` only, dropping `QUEUED`. The previous version used a separate 1h threshold for QUEUED, but the correct invariant is that the XRay viewer must never touch QUEUED at all. In the gpt-5.4 89-task run, opening the viewer mid-run promoted 3 unstarted QUEUED tasks to STALE — they were legitimately waiting for a Ray worker slot and would have been dispatched within minutes.

src/cube_harness/exp_runner.py
```diff
@@ -280,7 +280,7 @@ def _poll_ray(
         for task_ref in done:
             traj_id = ref_to_traj_id[task_ref]
             try:
-                traj: Trajectory = ray.get(task_ref)
+                traj: Trajectory = ray.get(task_ref, timeout=step_timeout_s + cancel_grace_s)
```
recursix, line 283: Add timeout to `ray.get()` so `_poll_ray` cannot block forever. Without this, if a worker hangs during teardown (e.g. `eai job kill` issues a CLOSE_WAIT), the entire polling loop freezes at this line and `_kill_stale_workers` never fires again. Budget is `step_timeout_s + cancel_grace_s` — the same ceiling already used by the worker kill decision logic just above.

tests/test_xray_promote_ghost.py
```diff
@@ -0,0 +1,77 @@
...
+import time
+from pathlib import Path
+
+import pytest
+
+from cube_harness.analyze.xray_utils import _promote_ghost_episodes
+from cube_harness.episode_status import STATUS_FILENAME, EpisodeStatus
+
+
+def _write_status(ep_dir: Path, status: str, age_seconds: float) -> None:
```
recursix, line 19: **Why `_write_status` sets `last_heartbeat_at = started_at`**

QUEUED episodes are pre-flight: no worker has picked them up yet, so they never emit heartbeats. This helper simulates the worst case — a QUEUED episode with a heartbeat timestamp identical to its start time (i.e., as stale as it can possibly look to `_promote_ghost_episodes`). This makes the test maximally strict: if the guard fails, the task age alone would be enough to trigger a promotion, and we'd catch it.

tests/test_xray_promote_ghost.py
```diff
@@ -0,0 +1,77 @@
...
+        status=status,
+        task_id="test-task",
+        episode_id=0,
+        started_at=time.time() - age_seconds,
+        last_heartbeat_at=time.time() - age_seconds,
+    )
+    s.write(ep_dir / STATUS_FILENAME)
+
+
+def test_queued_episode_never_promoted(tmp_path: Path) -> None:
```
recursix, line 31: **Bug this test pins: XRay viewer silently killed QUEUED tasks during large batches**

**Symptom**: In a 89-task gpt-5.4 batch, opening the XRay viewer mid-run caused 3 tasks to disappear from results. Post-mortem showed those episodes had status QUEUED (legitimately waiting for a Ray worker slot) when the viewer refreshed. `_promote_ghost_episodes` had a blanket guard `status not in ("RUNNING", "QUEUED")` — but that guard was written the wrong way: it skipped promotion only for those two statuses, meaning QUEUED was NOT skipped — it was promoted to STALE. The 1-hour threshold was easily exceeded for tasks at the back of a 4-worker queue.

**The fix**: promote only `RUNNING` episodes. QUEUED episodes are owned by the runner, not the viewer.

**Certainty**: High. This test directly exercises the updated predicate with a 99999-second-old QUEUED episode and asserts status is unchanged.

tests/test_xray_promote_ghost.py
```diff
@@ -0,0 +1,77 @@
...
+    _write_status(ep, "QUEUED", age_seconds=99_999)
+
+    _promote_ghost_episodes(tmp_path)
+
+    after = EpisodeStatus.read(ep / STATUS_FILENAME)
+    assert after is not None
+    assert after.status == "QUEUED"
+
+
+def test_old_running_episode_promoted_to_stale(tmp_path: Path) -> None:
```
recursix, line 43: **Regression guard: RUNNING promotion must still work**

The QUEUED fix must not accidentally disable the original feature. An old RUNNING episode (heartbeat gone silent for 99999s) represents a genuinely hung worker — it should be promoted to STALE so the viewer reflects reality and the runner can reassign or skip it. This test ensures the happy path is preserved.

tests/test_xray_promote_ghost.py
```diff
@@ -0,0 +1,77 @@
...
+    _write_status(ep, "RUNNING", age_seconds=30)
+
+    _promote_ghost_episodes(tmp_path)
+
+    after = EpisodeStatus.read(ep / STATUS_FILENAME)
+    assert after is not None
+    assert after.status == "RUNNING"
+
+
+@pytest.mark.parametrize("terminal_status", ["COMPLETED", "FAILED", "CANCELLED", "STALE"])
```
recursix, line 67: **Guard against double-promotion of already-resolved episodes**

Parametrized over all terminal statuses. A COMPLETED or FAILED episode should never be touched by `_promote_ghost_episodes`, regardless of age — the status is final. Without this check, a refactor could accidentally re-open a closed episode if the predicate logic shifts.

PR #376, opened 2026-05-12
Fix: Task class loading for workarena L2 & L3 

younik:
Those classes are created at runtime, so task class loading is failing for those


conte91 (approved):
PLUR, thanks!

PR #404, opened 2026-05-16
docs(pi): auto-fix provenance + fix methodology

recursix:
## Summary

Documentation/skill layer of a methodology to keep **PI-produced fixes trustable without a deep contextual review**, and to surface design rot instead of accreting narrow band-aids. Motivated by the tbench2 PI session: subtle, contextual bug-fix PRs are hard to review, the tendency is to rubber-stamp, and the codebase risks becoming a pile of overfit patches.

Core idea: **relocate trust** — a fix PR must carry its own generalization proof (the *Fix Dossier*) plus an independent adversarial audit of that proof, so the reviewer reads the proof, not the diff. In the autonomous loop there is no feedback channel to other users, so the durable, context-stamped **auto-fix provenance registry** *is* the cross-user signal.

### What's in (docs/skills only — no code)

- **`docs/auto-fix.md`** — authoritative spec: L0–L3 depth taxonomy; Fix Dossier; fix-audit; GitHub-issue-as-id provenance (one-line markers + module-bottom footnote with a machine/prose split + context fingerprint); two-tier rot detection (Tier-1 deterministic lint, Tier-2 semantic agent); **acknowledgement-not-prevention** (drift flags, never hard-blocks — else markers go radioactive); accretion (`grep -c`) → mandatory refactor proposal.
- **`templates/fix_dossier.md`** — the fix PR body template.
- **principal-investigator `SKILL.md`** — concise methodology summary linking the spec; dossier added to templates.
- **`AGENTS.md`/`CLAUDE.md`** — 5-line pointer under *Code review* (kept short to avoid context bloat).
- **`review-rules.md`** — **AF-001**, the Tier-2 rot check `/cube-review` runs.

### Policy highlights

- Nothing blocks the loop: L2/L3 (symptom-of-design) still ship a temp PR **and** open a kept-open `design-debt` issue + openspec stub. The human's queue becomes the design-debt backlog, not individual patches.
- Regression tests assert the **invariant**, not the reproduction.
- Generalization is proven via static blast-radius (on the *target* branch) + adversarial self-construction + registry lookup — never by polling users (impossible in the loop).

### Deferred to the next increment (code)

1. `analyze/investigator/use_cases/fix_audit/` recipe (reuses the #402 audit machinery).
2. `scripts/auto_fix_lint.py` — Tier-1 deterministic checks, wired into `make lint`.
3. Retrofit #7/#10/#11 as the first real Dossiers to pressure-test the template.

This doc layer is reviewable on its own and is the contract the above will implement against.

## Test plan

- [x] Docs/markdown only — no code paths touched; nothing to unit-test
- [x] Relative links resolve (`docs/auto-fix.md` from AGENTS.md root and `../../../docs/auto-fix.md` from the skill dir)
- [ ] Reviewer: read `docs/auto-fix.md` as the spec; confirm the policy (esp. acknowledgement-not-prevention, L2/L3 ship-don't-block) matches intent before the code increment builds on it


recursix (commented):
docs is for humans to read. specs is for coding agent. So there should be something in specs. Maybe more detailed. But in docs there should be some lightweight explanation of what are these auto-fix things

The auto-fix note should also point to the PR that fixed it. 

If gh is not available, auto-cube is not working any way so users should not use it. No need for fallback plan

PR #461, opened 2026-05-27
fix(storage): use mode="json" in model_dump to handle Decimal

conte91:
Trajectory metadata can contain Decimal values (e.g. from LLM usage costs) which json.dumps cannot serialize. model_dump(mode="json") coerces them to floats.

One-line fix in FileStorage.save_trajectory.


recursix (approved):
LGTM. Confirmed it fixes a real json.dumps crash on non-JSON-native metadata; regression test added. Clarifications on serialized type (Decimal→string) and the broader trigger are description-only and noted in a comment.

PR #462, opened 2026-05-27
fix(eval_log): coerce sample_id to str for integer task_ids

conte91:
Some benchmarks (e.g. WebArena) use integer task IDs. EpisodeRecord.sample_id is typed as str, so coerce at the assignment site to avoid type mismatches downstream.

One-line fix in EpisodeRecord.from_trajectory.


recursix (approved):
Clean one-line fix: coercing sample_id to str handles integer task_ids correctly. CI green.

PR #464, opened 2026-05-27
fix(webarena): reddit rate-limit relaxation + launch healthcheck diagnostics

conte91:
- **Reddit:** relax Postmill rate limits (SubmissionData / CommentData / UserData)
  + cache clear so the agent isn't throttled mid-task. Single source in
  `scripts/reddit_relax_limits.sh`, injected into both the reddit-only and
  all-sites launch paths from `resources.py` (`_REDDIT_RELAX`) so the two can't drift.
- **Diagnostics:** on healthcheck failure, dump container status, logs, and (for
  wikipedia) `/data` + kiwix/env-control versions, to speed up triage.

> **Note:** the kiwix-serve wikipedia migration originally bundled here was split
> into #507 (it needs a volume-population fix + a faithfulness check). This PR now
> contains only the reddit + diagnostics hardening.

### DIFF
--- a/cubes/webarena-verified/src/webarena_verified_cube/resources.py
+++ b/cubes/webarena-verified/src/webarena_verified_cube/resources.py
@@ -27,6 +27,13 @@ def _script(name: str) -> str:
     return (_SCRIPTS_DIR / name).read_text()
 
 
+# Reddit rate-limit relaxation, appended to every launch path that starts the
+# reddit container (reddit alone, or all sites together). Single source so the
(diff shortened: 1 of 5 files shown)


younik (approved):
> any chance you know something about the wiki container?

I don't :(

But this is the WebArena one, so I guess it should be fine?

I think it is worth opening an issue to https://github.com/ServiceNow/webarena-verified, explaining your issue with the Wiki image (or is it https://github.com/ServiceNow/webarena-verified/issues/38 ?)

PR #465, opened 2026-05-27
fix(workarena): seed generator, task class loading, infeasible tool key

conte91:
- Seed generator: L2/L3 tasks now get [meta_seed] directly instead of relying on get_all_tasks_agents curriculum (which varies with meta_seed and produced empty seed lists for some task types). L1 unchanged.
- Infeasible tool: use "message" key instead of "content" to match WorkArena validator expectation.
- Regenerated task_metadata.json (333 → 715 tasks) from canonical ALL_COMPOSITIONAL_TASKS_L2/L3 lists.


younik (approved):
PLUR!

PR #478, opened 2026-06-03
feat(rl): add Ray-backed RL rollout service and MiniWoB trainer integration recipes

Masseeh:
## Summary

This PR adds an RL rollout-service path to `cube-harness`, allowing it to run as a standalone rollout producer for external RL trainers (related issue #131).

The intended split is:

- `cube-harness` owns rollout generation.
- External RL trainers own training, replay buffers, optimization, persistence policy, and long-term storage.
- Rollout episodes emit structured events as they run.
- Trainers consume events asynchronously and reconstruct trajectories from event metadata.

Design references:

- Proposal: `openspec/changes/feat-rl/proposal.md`
- Deltas: `openspec/changes/feat-rl/deltas.md`
- RL spec: `openspec/specs/rl/spec.md`
- Related runtime hooks:
  - `openspec/specs/episode/spec.md`
  - `openspec/specs/storage/spec.md`
  - `openspec/specs/llm/spec.md`

### RL rollout service

- Added `src/cube_harness/rl/` as the rollout-service layer.
- Added an HTTP service for rollout submission, SSE event streaming, event acknowledgement, cancellation, health/stats, and task-config discovery via `/task-configs`.
- Added Ray-backed rollout execution with stateless workers.
- Added local rollout execution for debugging.
- Added request correlation metadata across events: `request_id`, `trajectory_id`, `task_id`, `group_id`, `rollout_index`, and per-request `event_index`.
- Added terminal events for rollout, worker, Ray, and training-data failures.

### Event streaming and reconstruction

- RL consumes the canonical `cube-harness` event stream through `RLEventSink`.
- External trainers can reconstruct trajectories from streamed events.
- Events include enough metadata for grouping, replay-by-offset, acking, and partial trajectory updates.
- Trainable LLM call events include prompt token IDs, completion token IDs, and logprobs.

### Rollout LLM support

- Added rollout-specific LLM config and response handling in `cube_harness.rl.llm`.
- Captures training-critical metadata: `prompt_token_ids`, `completion_token_ids`, and completion logprobs.
- Fails closed when trainable rollout calls are missing required token/logprob data.
- Keeps LiteLLM as the transport, while currently relying on vLLM/OpenAI-compatible token-ID behavior. Other inference engines need explicit validation.

### Persistence model

- Normal rollout-service operation avoids full disk-backed rollout artifacts.
- Disk persistence is opt-in debug/replay behavior via `persist_rollout=True`.
- Trainers are expected to consume the event stream and decide what to persist.

### MiniWoB examples

Added two RL recipes:

- `recipes/rl/hello_miniwob_service.py`
  - Reference service-style trainer integration.
  - Starts the rollout HTTP service.
  - Queries `/task-configs`.
  - Submits rollouts.
  - Consumes SSE events asynchronously.
  - Reconstructs trajectories.
  - Groups rollouts per task config.
  - Computes group rewards.
  - Writes token-ID SFT JSONL examples.

- `recipes/rl/hello_miniwob_local.py`
  - Direct in-process rollout debugging.
  - No HTTP server.
  - Prints rollout events as they are generated.

The TIR cube and TIR recipe were moved out of this PR and preserved separately on `rl/tir`.

## Reference Usage

Service-style trainer workflow:

```bash
CUBE_HARNESS_LLM_BASE_URL=http://127.0.0.1:8000 \
CUBE_HARNESS_MODEL="MODEL-NAME" \
CUBE_HARNESS_TOKENIZER_NAME="PATH-TO-TOKENIZER" \
uv run recipes/rl/hello_miniwob_service.py --num-rollouts 10 --num-groups 2
```

Local debugging workflow:

```bash
CUBE_HARNESS_LLM_BASE_URL=http://127.0.0.1:8000 \
CUBE_HARNESS_MODEL="MODEL-NAME" \
CUBE_HARNESS_TOKENIZER_NAME="PATH-TO-TOKENIZER" \
uv run recipes/rl/hello_miniwob_local.py
```

Deterministic smoke:

```bash
uv run scripts/smoke/rl_mock_multiturn_service.py --turns 2
```

## Testing

Focused coverage includes:

- rollout service health and request handling,
- task-config discovery,
- event ordering and reconstruction metadata,
- duplicate rollout request handling,
- Ray/local execution behavior,
- cancellation and terminal failure events,
- debug persistence opt-in,
- prompt/completion token ID capture,
- non-trainable rollout detection,
- rollout LLM training metadata capture.

Validation run:

```bash
uv run pytest tests/test_rollout_service.py tests/test_default_agent_run.py tests/test_monitored_tool_compat.py tests/test_storage_event_layout.py
uv run scripts/smoke/rl_mock_multiturn_service.py --turns 2
make lint
```

Latest local result:

- `50 passed, 3 warnings`
- RL smoke passed
- lint passed

## Notes

This PR intentionally does not implement a full RL trainer. It makes `cube-harness` a high-throughput rollout producer. The MiniWoB service recipe is the reference integration pattern for external trainers such as `PipelineRL`.


src/cube_harness/streamer.py
```diff
@@ -175,6 +178,8 @@ def emit(self, te: TrajectoryEvent) -> str:
                 sink.save_event(te, self.trajectory_id)
             except Exception:
                 logger.exception("EventStreamer sink %r raised; continuing.", sink)
+                # TODO: RL rollout need to raise so the worker fails and the executor can emit an error terminal.
```
recursix: **Fail-fast publishing isn't actually wired** — `emit` swallows every sink exception (this TODO), so `RLEventSink._publish`'s `raise` never propagates. The executor's `has_terminal` fallback covers a publisher that's fully down at the end, but a *transient mid-rollout* publish failure produces a terminal with `rollout_status="completed"` over a silently partial event stream (only `rollout_valid=False` hints at it) — violating the "a trajectory is never silently partial" intent. The commented-out `test_event_streamer_can_raise_required_sink_failure` acknowledges this.

Suggestion: a per-sink `required` marker on the streamer — re-raise for required sinks, keep best-effort for the rest — and register the RL sink as required. Then `test_rl_sink_publisher_failure_is_recorded` should drive `streamer.emit(...)` rather than `sink.save_event(...)` so it tests the production path. Your call on the mechanism; happy to discuss.

src/cube_harness/rl/task_runner.py
```diff
@@ -0,0 +1,63 @@
...
+        self.payload = payload
+        self.publisher_handle = publisher_handle
+        self.request = dict(payload["request"])
+        self.request_id = str(self.request["request_id"])
+        self.task_config = payload["task_config"]
+        self.agent_config = copy.deepcopy(payload["agent_config"])
+        self.output_dir = Path(payload["output_dir"])
+        self.episode_id = int(self.request.get("rollout_index") or 0)
+        self.task_id = str(self.request["task_id"])
+        self.trajectory_id = trajectory_log_id(self.task_id, self.episode_id)
```
recursix: **`trajectory_id` collides across groups.** It's built from `(task_id, rollout_index)` only — two groups rolling the same task with the same `rollout_index` get the same `trajectory_id`. The docs tell trainers to reconstruct trajectories keyed on it, and in `persist_rollout=True` mode the second episode hits `FileStorage.save_metadata`'s `FileExistsError`. `request_id` is the actual unique key — suggest deriving `trajectory_id` from it (or making them equal) and keeping task/index as metadata. Since this is the wire contract PipelineRL will integrate against, better to fix before it ossifies.

src/cube_harness/rl/events.py
```diff
@@ -0,0 +1,95 @@
...
+    if isinstance(value, dict):
+        return {str(k): _jsonable(v) for k, v in value.items()}
+    if hasattr(value, "model_dump"):
+        return _jsonable(value.model_dump(exclude_none=True))
+    if hasattr(value, "dict"):
+        return _jsonable(value.dict(exclude_none=True))
+    return str(value)
+
+
+class EventContext(BaseModel):
```
recursix, line 39: **Add a nullable `agent_id` to the event context now.** The multi-agent episode work (#497, companion to cube-standard #214) tags every `ToolCallEvent`/`LLMCallEvent` with an `agent_id` — one episode will produce per-agent event slices, and for RL the trainable unit becomes the per-agent sequence (think self-play groups). Adding `agent_id: str | None = None` to the payload envelope today is one field; adding it after PipelineRL integrates is a wire-contract break.

src/cube_harness/rl/sink.py
```diff
@@ -0,0 +1,169 @@
...
+    return _UNSAFE_PATH_CHARS.sub("_", value) or "unknown"
+
+
+class EventSinkConfig(BaseModel):
+    max_hot_events: int = 10000
+    persist_events_dir: Path | None = None
+    event_publish_timeout_s: float = 30.0
+
+
+class EventSink:
```
recursix: **Naming: this `EventSink` collides with the `EventSink` Protocol in `cube_harness.streamer`** (the feat-rl proposal itself notes the clash). They're different concepts — the streamer one is "consumes trajectory events", this one is a transport buffer (offsets/acks/spill/SSE). Suggest `EventBuffer` or `EventPublisher` for this class before downstream code starts importing both.

src/cube_harness/rl/trajectory_sink.py
```diff
@@ -0,0 +1,208 @@
...
+                "llm_call_tag": call.tag if call is not None else "",
+            }
+        if event.error is not None:
+            self.error = dump_for_event(event.error)
+            self.status = "llm_error"
+        if trainable:
+            self.rollout_trainable = True
+
+        payload = rollout_event_payload(self.ctx, "llm_call", self.event_idx)
+        payload.update(
```
recursix: **Consider envelope + `model_dump()` instead of per-field copies.** These payload builders hand-copy each canonical event field, so any field later added to `LLMCallEvent`/`ToolCallEvent` (e.g. the `agent_id` coming with #497) is silently dropped from the wire unless someone remembers to add it here too. A generic shape — context envelope (ids/offset/type/timestamp) + `dump_for_event(te.output)` — would auto-inherit core event evolution and delete ~50 lines. Not blocking, but it changes the payload shape, so better decided before trainers integrate.

src/cube_harness/streamer.py
Masseeh: Addressed in [7493e8e](https://github.com/The-AI-Alliance/cube-harness/commit/7493e8e810f51b9c3daef92c1ee19d16f6793cf2)

Added a required raise_on_emit_error flag to EventSink. Sinks that set this flag will now re-raise emit errors, allowing required sinks to fail the worker and propagate an error terminal. Other sinks remain best-effort: errors are logged and emission continues for the remaining sinks.

recursix (approved):
## Approved ✅

Review complete (multi-agent review pass + manual passes with Claude; all findings adversarially validated). The architecture is right: a rollout is a normal `Episode` with an `RLEventSink` on the existing sink seam — no episode-loop fork, no disk on the hot path, trainer owns policy/reward/optimization.

**Landed during review (on the branch):**
- BudgetExceeded rollouts keep their terminal `task.evaluate()` reward (`bb318383`)
- Required-sink fail-fast marker — thanks for the quick turnaround on this one (`7493e8e8`)
- Typer CLIs + `ch-rollout` entry-point wrapper; single token-limit param; sanitized JSONL spill paths (`91e3a050`)
- Synced with dev (the agents diff is now just the llm_config typing), re-pinned the cube-standard lock to current dev tip, excluded `recipes/rl/` from the Experiment recipe guard (`5611d583`, `28e0cf03`)
- `llm_config: SerializeAsAny[BaseLLMConfig]` — single open type instead of the union; removes the agents → `cube_harness.rl` import (`1466d36d`)

**Non-blocking, but please consider before/at merge** (inline comments have details):
1. `trajectory_id` derived from `request_id` (cross-group collision — this is wire contract, cheapest to fix now)
2. Nullable `agent_id` in the event context (multi-agent #497 is coming)
3. Rename the rl `EventSink` (collides with the streamer Protocol)
4. Envelope + `model_dump()` payloads vs per-field copies

**Follow-ups we'll track separately:**
- Cut cube-standard `0.1.0rc9` → drop the git source pin + re-pin (same cycle as `279814d3`)
- Mark the engine/executor/service layer experimental in the spec; convergence RFC with `exp_runner` (shared Ray scheduling, heartbeat sweeps, retries) once PipelineRL gives real usage data — co-authored, when you're ready

Nice work, Masih — excited to see PipelineRL run against this.

src/cube_harness/rl/sink.py
Masseeh: Addressed in 03829ed6f104a75427a7999dc90a0260b039a2fe and d2a4985a8f5f548d35dfa079090d0a6038ea8f8f

- changed the EventSink and related classes to EventPublisher
- changed the module name from sink.py to event_publisher.py

src/cube_harness/rl/events.py
Masseeh: Addressed in 8fec4e3c99705900fa3e1619165407d57acc30bf

src/cube_harness/rl/trajectory_sink.py
Masseeh: Addressed in 15114e4a8c69f67527e7c6f6d5b0aa906247f1db

Implemented the envelope/body RL payload shape.

  Changed src/cube_harness/rl/trajectory_sink.py so trajectory-derived rollout events now publish:

  - top-level context envelope: type, event_index, request/task ids, timestamp, etc.
  - event: dump_for_event(te.output) canonical event body
  - rl: rollout-only annotations like trainable, llm_call_index, tool_call_index, state_ref
  - trajectory_event: start_time / end_time for tool-call timing

  Accepted and terminal events stay flat because they are rollout control events, not canonical trajectory events.

src/cube_harness/rl/task_runner.py
Masseeh: Addressed in b709e183ae26a24801afcc00999d7bb154cb2b5a

  - RL now sets trajectory_id == request_id.
  - Episode supports an optional trajectory_id override while normal experiments still use {task_id}_ep{episode_id}.
  - Persisted rollout debug artifacts now land under <output_dir>/<request_id>/episodes/<request_id>/.
  - Updated storage config saving to use the resolved trajectory id.
  - Updated RL tests, smoke expectations, recipes, and OpenSpec docs.
  - Added regression coverage for same task_id + same rollout_index across different groups.

PR #525, opened 2026-07-20
Added knows-cube — 110 Google Workspace authoring tasks

farhanishmam:
Ports KNOWS — 22 families × 5 instances across Google Docs (25), Sheets (45)
and Slides (40). The agent drives a real browser against `docs.google.com`;
grading calls the upstream KNOWS evaluators.

Depends-on: cube-standard/dev

## Approach

Follows the `webarena-verified` pattern: depends on the KNOWS eval/task layer
directly, with stock `browsergym-core` from PyPI. No BrowserGym-Knows fork, no
`PYTHONPATH` injection, no `AgentLab-Knows` submodule — KNOWS needs only
`register_task` and `AbstractBrowserTask`.

Evaluator loading is reimplemented rather than reused, specifically so that
upstream's `_load_evaluator` `.env` scraping never runs. Upstream pulls `.env`
files from two directories above its repo root into `os.environ`; the KNOWS
repos ship committed live API keys.

## Correctness work beyond the straight port

**Derived evaluator paths.** Upstream's paths are correct *in upstream's
layout* — `get_base_path()` returns `os.getcwd()`, so they resolve when run
from the Agent-Benchmark repo root. This cube consumes the wheel, which has no
`src/` prefix. Rewriting only the named constants left 47 derived ones (e.g.
`GOLD_IMAGE_ORIGINAL`, `TASK_MD_PATH`) across **21 of 110 instances** pointing
at a nonexistent tree. Evaluators guard those reads with `os.path.exists()`, so
a stale path *silently skipped a scoring step* — a reward drop with no
`evaluation_error` to distinguish it from an agent failure.

**Browsing history hook.** Recorded in `Task._post_action` (cube-standard's
per-action hook) rather than `obs_postprocess`, which fires once per *step* and
*after* `evaluate()`. The latter under-counted multi-action steps in the
`browsing_history` that 94 of the 110 evaluators grade against. This is why the
`cube-standard` floor is rc11.

**Terminal eval record.** `evaluate()` caches and returns copies of its info
dict; `Task.step` writes `info["profiling"]` into the returned object, which
otherwise mutated the cached record persisted for the episode.

**Debug config serialization.** `tool_config` is `SerializeAsAny[ToolConfig]`,
not `Any` — re-annotating as `Any` dropped both validation and the polymorphic
serializer, so the config round-tripped through Ray/storage as a plain dict.

## Verification

- 33 unit tests; `cube test knows-cube` → 2/2, reward 1.0
- Full `cube_harness` run: `Experiment → Episode → Trajectory → FileStorage`
  with a scripted agent, confirming the eval record persists clean
- `ruff check` + `ruff format` clean

**Not verified against real LLMs.** Please don't read the green checks as
"this benchmark is known to work end-to-end." A real episode needs a Google
service account + paid Gemini credentials, and `cube-browser-playwright`
cannot yet supply an authenticated browser session (no `user_data_dir`, no
`storage_state`; `launch_persistent_context` has no such parameter). See
README "Blocking gap".

## Open design questions for review

- `abstract_description` currently holds the full `task.md` (median 956 chars
  vs ≤460 across every sibling cube). cube-standard's docstring says heavy
  problem statements don't belong there. Fixing means regenerating the 213 KB
  metadata — happy to do it if you want a one-line summary instead.
- A deliberate agent stop is recorded as `episode_end="truncated"`, so it's
  indistinguishable from budget exhaustion in any analysis partitioning on
  that key. Fix is either `accept_agent_stop = False` or a fourth label.


recursix (approved):
Awesome Thank you for the addition :)

